PR #1, opened 2023-10-13
Add a tool to AOT compile TCP snippets into CPU binaries

sanjoy:
For now I think it will help in writing execution tests.  Longer term we may find other uses for it.

Here are the changes I made:

1. I added a `-lower-tcp-to-llvm` pass pipeline.
2. Created a genrule based macro that invokes `tcp-opt`, `mlir-translate` & `llc` to generate the CPU binary.
3. Added some tests.

### DIFF
--- a/lib/Pipeline/Pipeline.cpp
+++ b/lib/Pipeline/Pipeline.cpp
@@ -9,8 +9,8 @@
 
 #include "Pipeline/Pipeline.h"
 
-#include "Conversion/TcpToLinalg/TcpToLinalg.h"
 #include "Conversion/TcpToArith/TcpToArith.h"
+#include "Conversion/TcpToLinalg/TcpToLinalg.h"
 
 #include "mlir/Conversion/FuncToLLVM/ConvertFuncToLLVMPass.h"
 #include "mlir/Conversion/MathToLLVM/MathToLLVM.h"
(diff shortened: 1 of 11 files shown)


test/BUILD
```diff
@@ -43,3 +44,15 @@ filegroup(
     )
     for src in glob(["**/*.mlir"])
 ]
+
+aot_compile("basic_tcp_ops", "basic_tcp_ops.mlir")
```
sjain-stanford, line 48: readability nit: specify as keyword args
```
aot_compile(
    name = "basic_tcp_ops",
    tcp_source = "basic_tcp_ops.mlir",
)
```

sjain-stanford (approved):
LGTM, this should help with some integration test coverage for TCP ops. 

Btw it would be nice to have some documentation / contributor guideline for someone adding new ops to TCP (without this it'd be easy to forget adding them to this AOT compile flow).

lib/Pipeline/Pipeline.cpp
```diff
@@ -0,0 +1,57 @@
...
+  pm.addPass(createLinalgBufferizePass());
+  pm.addNestedPass<func::FuncOp>(createConvertLinalgToLoopsPass());
+  pm.addPass(createConvertSCFToCFPass());
+
+  pm.addPass(memref::createExpandStridedMetadataPass());
+  pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+  pm.addPass(createFinalizeMemRefToLLVMConversionPass());
+  pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+  pm.addPass(createConvertMathToLibmPass());
+  pm.addPass(createConvertMathToLLVMPass());
```
navahgar, line 48: One issue I have noticed with doing `math-to-libm` before `math-to-llvm` is that the former marks `math` dialect as illegal, so it expects no `math` ops to exist after the pass. But also, it does not handle all the `math` ops.

We should either update the upstream code to remove the illegal part or do it after `math-to-llvm` (which is what I had to do in our internal pipeline to fix this issue - you may not see it yet since it hasn't landed).

test/BUILD
sanjoy: Done.

PR #4, opened 2023-10-14
[NFC] Create Dialects, move IR and Transforms within it

sjain-stanford:
Move IR and Transforms within Dialect to match with MLIR / Torch-MLIR convention. This is a non-functional change.


lib/TorchConversion/Transforms/VerifyTcpBackendContract.cpp
```diff
@@ -7,10 +7,10 @@
 //
 //===----------------------------------------------------------------------===//
 
-#include "PassDetail.h"
+#include "../PassDetail.h"
 
-#include "IR/TcpDialect.h"
-#include "IR/TcpOps.h"
+#include "Dialect/IR/TcpDialect.h"
+#include "Dialect/IR/TcpOps.h"
```
sjain-stanford, line 13: These are not being tested at the moment, I plan to send a follow-on to add `TorchBackendToTcpBackendPipeline` along with LIT tests that should cover this pass.

sanjoy (approved):

PR #6, opened 2023-10-14
Update README

sjain-stanford:
Add CI status badge.
Add `mlir-tcp` discord links.
Add developer guide section and include build/test/lint steps.

Here's how it would look like: https://github.com/cruise-automation/mlir-tcp/blob/f1eaf57340a302c37f420258cce5f7f42822785b/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,30 +5,29 @@ Mid-level intermediate representation for machine learning programs.
 
 ![Bazel Build](https://github.com/cruise-automation/mlir-tcp/actions/workflows/bazelBuildAndTestTcp.yml/badge.svg)
 
+:construction: **This project is under active development (WIP).**
+
 ## Project Communication
 
 - For general discussion use `#mlir-tcp` channel on the [LLVM Discord](https://discord.gg/xS7Z362)
-- File [issues](https://github.com/cruise-automation/mlir-tcp/issues) on GitHub
+- For feature request or bug report file a detailed [issue on GitHub](https://github.com/cruise-automation/mlir-tcp/issues)
 
 ## Developer Guide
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,10 +1,43 @@
...
+- File [issues](https://github.com/cruise-automation/mlir-tcp/issues) on GitHub
+
+## Developer Guide
+
+To build TCP using Bazel, follow these steps:
+
+1. For a quick start, launch an interactive docker container with clang (and lld) pre-installed:
+
+```shell
+./docker/run_docker.sh
```
sanjoy, line 20: Do we really need this for TCP? Things were more complicated for Torch-MLIR, but for TCP we just need clang.

README.md
```diff
@@ -1,10 +1,43 @@
 Tensor Compute Primitives
 =========================
 
-Mid level intermediate representation for machine learning programs.
+Mid-level intermediate representation for machine learning programs.
```
sanjoy, line 4: Add a note that the project is WIP.

sanjoy (approved):
Minor comments, but overall LGTM.

README.md
sjain-stanford: You're correct that this is an optional step (marked it as `Optional` now). I'd recommended it because we use the same minimal docker to source the GHA runners for CI (for reproducibility reasons and cruise restrictions on installing stuff on GHA runners that are not from GitHub marketplace) and historically (with Torch-MLIR) we've seen issues where CI failures were hard to reproduce locally for our users (due to environment issues, clang versions). Having a docker enforces some form of hermeticity. Another advantage is in case we grow to take more dependencies this is more future proof.

PR #14, opened 2023-10-26
Create TcpTypes.td and move type definitions there

sjain-stanford:
As titled.

### DIFF
--- a/include/mlir-tcp/Dialect/IR/TcpBase.td
+++ b/include/mlir-tcp/Dialect/IR/TcpBase.td
@@ -11,8 +11,11 @@
 #define TCP_BASE
 
 include "mlir/IR/OpBase.td"
-
 include "mlir/Interfaces/SideEffectInterfaces.td"
+
+//===----------------------------------------------------------------------===//
+// Tcp Dialect Base
+//===----------------------------------------------------------------------===//
 
 def Tcp_Dialect : Dialect {
   let name = "tcp";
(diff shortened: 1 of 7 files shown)


include/mlir-tcp/Dialect/IR/TcpBase.td
```diff
@@ -97,4 +59,12 @@ class Tcp_BinaryElementwiseOp<string mnemonic, list<Trait> traits = []> :
...
 }
 
+//===----------------------------------------------------------------------===//
+// Tcp Types Base
+//===----------------------------------------------------------------------===//
+
+class Tcp_Type<string name, string typeMnemonic> : TypeDef<Tcp_Dialect, name> {
+    let mnemonic = typeMnemonic;
+}
+
```
navahgar, line 69: Why not move this also to `TcpTypes.td`?

navahgar (approved):
LGTM

include/mlir-tcp/Dialect/IR/TcpBase.td
sjain-stanford: No particular reason other than maintaining the pattern we followed for `Tcp_Dialect` and `Tcp_Op`. Upstream dialects don't follow any particular pattern either - e.g. Tosa uses a separate `TosaOpsBase.td` and `TosaTypesBase.td` whereas GPU dialect uses the same `GPUBase.td` for dialect, enum, type, interfaces and attrs.

PR #15, opened 2023-10-26
[Dialect] Traits and type parser/printer fixes to unblock downstream integrations

sjain-stanford:
- [x] Remove ParentOneOf trait from tcp group ops
  - No longer a constraint; fixes `error: 'tcp.isolated_group' op expects parent op to be one of 'func.func, tcp.group, tcp.isolated_group'`
- [x] Use default type parser/printer for TcpDialect
  - Used downstream; fixes `error: dialect 'tcp' provides no type parsing hook`
- [x] Add `Tcp_IndexArrayType` definition
  - Fixes `ld.lld: error: undefined symbol: mlir::tcp::TcpDialect::printType(mlir::Type, mlir::DialectAsmPrinter&) const` `referenced by TcpDialect.cpp`
- [x] Minor: rename pass registration

### DIFF
--- a/include/mlir-tcp/Dialect/IR/TcpTypes.td
+++ b/include/mlir-tcp/Dialect/IR/TcpTypes.td
@@ -16,11 +16,19 @@ include "mlir/IR/DialectBase.td"
 include "mlir-tcp/Dialect/IR/TcpBase.td"
 
 //===----------------------------------------------------------------------===//
-// Tcp Type Definitions.
+// Tcp Type Definitions
+//===----------------------------------------------------------------------===//
+
+def Tcp_IndexArrayType : Tcp_Type<"IndexArray", "index_array"> {
+    let summary = "IndexArray TCP type, to holds a list of index builtin type to represent shape";
+}
+
+//===----------------------------------------------------------------------===//
+// Tcp Quantized Types
 //===----------------------------------------------------------------------===//
 
 // The base class of a quantized type.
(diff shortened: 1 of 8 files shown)


include/mlir-tcp/Dialect/IR/TcpTypes.td
```diff
@@ -16,11 +16,19 @@ include "mlir/IR/DialectBase.td"
...
 
 //===----------------------------------------------------------------------===//
-// Tcp Type Definitions.
+// Tcp Type Definitions
+//===----------------------------------------------------------------------===//
+
+def Tcp_IndexArrayType : Tcp_Type<"IndexArray", "index_array"> {
+    let summary = "IndexArray TCP type, to holds a list of index builtin type to represent shape";
+}
+
```
navahgar, line 25: very minor nit: It might be better to have the basic types, `Tcp_Scalar`, `Tcp_Tensor` before we define `IndexArrayType`. The same would apply to the `Tcp_QuantizedType` as well. So, may be we can do it later.

navahgar (approved):
LGTM

include/mlir-tcp/Dialect/IR/TcpTypes.td
sjain-stanford: `Tcp_Scalar` depends on `Tcp_QuantizedInt` so it has to follow quantized types. However I've moved Tcp_IndexArrayType to the end.

PR #16, opened 2023-11-02
Allow partial `TorchToTcp` conversions using a whitelist

sjain-stanford:
Add pass option to TorchToTcp accepting a set of op name strings as whitelist (for selective conversion). This allows flexibility with phase ordering of conversion passes in downstream pipelines.

### DIFF
--- a/lib/Conversion/TorchToTcp/Elementwise.cpp
+++ b/lib/Conversion/TorchToTcp/Elementwise.cpp
@@ -652,15 +652,14 @@ void torch_to_tcp::populateElementwisePatternsAndLegality(
     TypeConverter &typeConverter, RewritePatternSet &patterns,
     ConversionTarget &target, const llvm::StringSet<> &convertTorchOpsSet) {
 
-#define INSERT_ATEN_ELEMENTWISE_OP_PATTERN(ConvertAtenOpPattern, AtenOp)       \
-  torch_to_tcp::addPatternIfOpInConvertTorchOpsSet<ConvertAtenOpPattern,       \
-                                                   AtenOp>(                    \
+#define INSERT_ATEN_ELEMENTWISE_OP_PATTERN(AtenOp)                             \
+  torch_to_tcp::addPatternIfOpInConvertTorchOpsSet<Convert##AtenOp, AtenOp>(   \
       typeConverter, patterns, target, convertTorchOpsSet)
-  INSERT_ATEN_ELEMENTWISE_OP_PATTERN(ConvertAtenToDtypeOp, AtenToDtypeOp);
-  INSERT_ATEN_ELEMENTWISE_OP_PATTERN(ConvertAtenClampOp, AtenClampOp);
-  INSERT_ATEN_ELEMENTWISE_OP_PATTERN(ConvertAtenReluOp, AtenReluOp);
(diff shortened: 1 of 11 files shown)


zezhang (approved):
Thanks for enabling this feature, LGTM!

lib/Conversion/TorchToTcp/TorchToTcp.cpp
```diff
@@ -49,6 +60,15 @@ class ConvertTorchToTcp : public ConvertTorchToTcpBase<ConvertTorchToTcp> {
 
   void runOnOperation() override {
     MLIRContext *context = &getContext();
+    RewritePatternSet patterns(context);
+
+    // The strings in the `convertTorchOps` ArrayRef don't exist during the call
```
sanjoy, line 65: `The strings in the `convertTorchOps` ArrayRef don't exist` is weirdly phrased IMO; is it correct to say that usually the default constructor is called which means `convertTorchOps` is usually unset.

lib/Pipeline/Pipeline.cpp
```diff
@@ -35,7 +35,10 @@
 using namespace mlir;
 
 static void createTorchBackendToTcpBackendPipeline(OpPassManager &pm) {
-  pm.addNestedPass<func::FuncOp>(tcp::createConvertTorchToTcpPass());
+  ArrayRef<std::string> emptyArrayRef;
+
+  pm.addNestedPass<func::FuncOp>(
+      tcp::createConvertTorchToTcpPass(emptyArrayRef));
```
sanjoy, line 41: Can this be `tcp::createConvertTorchToTcpPass({})`?

sanjoy (approved):

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -650,78 +650,59 @@ class ConvertAtenToDtypeOp : public OpConversionPattern<AtenToDtypeOp> {
...
-      typeConverter, context);
-
-  target.addIllegalOp<AtenBatchNormOp>();
-  patterns.add<ConvertAtenBatchNormOp>(typeConverter, context);
-
-  target.addIllegalOp<AtenAtan2Op>();
-  patterns.add<ConvertAtenAtan2Op>(typeConverter, context);
+    ConversionTarget &target, const llvm::StringSet<> &convertTorchOpsSet) {
+
+#define INSERT_ATEN_ELEMENTWISE_OP_PATTERN(ConvertAtenOpPattern, AtenOp)       \
```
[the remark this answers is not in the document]
sjain-stanford: Good suggestion, done. I think this was applicable to just `INSERT_ATEN_ELEMENTWISE_OP_PATTERN` macro in Elementwise.cpp and `INSERT_ATEN_MISC_OP_PATTERN` macro in Misc.cpp. PLMK if I missed any other occurrence and I'll address in a follow-on.

lib/Conversion/TorchToTcp/TorchToTcp.cpp
sjain-stanford: Rephrased as suggested.

lib/Pipeline/Pipeline.cpp
sjain-stanford: Done.

PR #17, opened 2023-11-03
Tcp Custom Op and Torch->Tcp conversions + lit tests

sjain-stanford:
Op definition for `tcp.custom_op` and lowerings for `aten.gather` and `aten.index.Tensor_hacked_twin`. 

This is useful to represent and lower custom/opaque operations that have tensor inputs and outputs, but 
1) lack a clear operational semantics, or 
2) have an operational semantic that isn't defined by Tcp (such as user defined custom operations), or
3) need to "flow through" the middleend because they target pre-written kernels available in the backend

As a follow-on, I'll look into making `torch -> tcp.custom_op` conversions automatic as much as possible, to minimize the overhead in adding individual conversion patterns manually. Then this could be the last step in a frontend lowering pipeline, to have all remaining `torch.aten` ops converted.


sanjoy (approved):

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include "llvm/ADT/StringSet.h"
+
+using namespace mlir;
+using namespace mlir::tcp;
+using namespace mlir::torch;
+using namespace mlir::torch::Torch;
+
+namespace {
+class ConvertAtenGatherOp : public OpConversionPattern<AtenGatherOp> {
```
navahgar, line 29: What if we add a `tcp.gather` op because we want to codegen that? Do you plan to provide some flag to control this, for existing Tcp ops?

IMO we should not have custom op transformations for ops that potentially have in TCP.

sjain-stanford: There are a few ways to support this: 
1. Explicit: Support the two paths in ConvertAtenGatherOp and control that from a pass option
1. Explicit: Keep them as two separate conversion patterns and control that from frontend pipeline (phase ordering + selective conversions)
1. Implicit: Lean towards codegen first (when available) and fallback to tcp.custom_op if not

sjain-stanford: To elaborate a bit more, here are the options to support this use-case:

**Option 1:**
Add a bool flag (say `enableCustomOpPatternsInstead`) to `ConvertTorchToTcp` constructor ([here](https://sourcegraph.com/github.com/cruise-automation/mlir-tcp/-/blob/lib/Conversion/TorchToTcp/TorchToTcp.cpp)), and selectively include either tcp op patterns or tcp custom op patterns like so:
```
if (enableCustomOpPatternsInstead) {
    torch_to_tcp::populateTcpCustomOpPatternsAndLegality(
        typeConverter, patterns, target, convertTorchOpsSet);
} else {
    torch_to_tcp::populateElementwisePatternsAndLegality(
        typeConverter, patterns, target, convertTorchOpsSet);

    torch_to_tcp::populateMiscPatternsAndLegality(typeConverter, patterns,
                                                  target, convertTorchOpsSet);

    torch_to_tcp::populateDataMovementPatternsAndLegality(
        typeConverter, patterns, target, convertTorchOpsSet);
}
```

**Option 2:**
Remove `torch_to_tcp::populateTcpCustomOpPatternsAndLegality` from `ConvertTorchToTcp` entirely, and create a separate `ConvertTorchToTcpCustomOps` conversion pass instead. We can then include `ConvertTorchToTcpCustomOps` after `ConvertTorchToTcp` in our pipeline (to prefer codegen over kernels) but also craft all combinations to override this easily.

To me both these options look pretty similar, but I am leaning towards 2 over 1 because:

- organizationally I like 2 more than 1 because it reduces the mental overhead on the pipeline stitching side a bit
- 1 is slightly non-idiomatic because we have to be careful about avoiding to populate two patterns that match the same torch source op (e.g. aten.gather). Keeping them as a separate pass prevents that from happening by design.
- 2 allows future extensions such as generic custom op conversion more easily

PLMK your thoughts, or if you have other ideas. I can send a PR for either options relatively quickly, but I wanted to check-in with you first.

sjain-stanford: Decided to go with Option 2 for now: https://github.com/cruise-automation/mlir-tcp/pull/19

PR #19, opened 2023-11-04
Separate conversion pass for `torch -> tcp.custom_op` to support both kernel and codegen approaches

sjain-stanford:
This allows switching between codegen vs kernels approaches, for torch ops that can lower to either tcp ops (codegen) or tcp.custom_op (kernel). Addresses https://github.com/cruise-automation/mlir-tcp/pull/17#discussion_r1382303711 using [Option 2](https://github.com/cruise-automation/mlir-tcp/pull/17#discussion_r1382436616).

### DIFF
--- a/include/mlir-tcp/Conversion/TorchToTcp/TorchToTcpCustomOp.h
+++ b/include/mlir-tcp/Conversion/TorchToTcp/TorchToTcpCustomOp.h
@@ -21,6 +21,7 @@ namespace tcp {
 
 std::unique_ptr<OperationPass<func::FuncOp>>
 createConvertTorchToTcpCustomOpPass();
+
 std::unique_ptr<OperationPass<func::FuncOp>>
 createConvertTorchToTcpCustomOpPass(
     llvm::ArrayRef<std::string> convertTorchOps);
--- a/include/mlir-tcp/Conversion/TorchToTcp/TorchToTcpCustomOp.h
+++ b/include/mlir-tcp/Conversion/TorchToTcp/TorchToTcpCustomOp.h
@@ -1,0 +1,29 @@
+//===------------------------------------------------------------*- C++ -*-===//
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// Also available under a BSD-style license. See LICENSE.
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "mlir/Dialect/Func/IR/FuncOps.h"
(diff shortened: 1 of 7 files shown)


include/mlir-tcp/Conversion/TorchToTcp/TorchToTcpCustomOp.h
```diff
@@ -0,0 +1,29 @@
...
+namespace mlir {
+
+#define GEN_PASS_DECL_CONVERTTORCHTOTCPCUSTOMOP
+#include "mlir-tcp/Conversion/Passes.h.inc"
+
+namespace tcp {
+
+std::unique_ptr<OperationPass<func::FuncOp>>
+createConvertTorchToTcpCustomOpPass();
+std::unique_ptr<OperationPass<func::FuncOp>>
```
sanjoy, line 24: Nit: add newline between function declarations.

lib/Conversion/TorchToTcp/TorchToTcp.cpp
```diff
@@ -97,6 +94,11 @@ class ConvertTorchToTcp : public ConvertTorchToTcpBase<ConvertTorchToTcp> {
 
 } // namespace
 
+std::unique_ptr<OperationPass<func::FuncOp>> createConvertTorchToTcpPass() {
+  llvm::ArrayRef<std::string> emptyArrayRef;
+  return std::make_unique<ConvertTorchToTcp>(/*convertTorchOps=*/emptyArrayRef);
```
sanjoy, line 99: Can this be `std::make_unique<ConvertTorchToTcp>({})`

sanjoy (approved):

lib/Conversion/TorchToTcp/TorchToTcp.cpp
sjain-stanford: I initially had that, but hit this compiler error:
```
lib/Conversion/TorchToTcp/TorchToTcp.cpp:99:10: error: no matching function for call to 'make_unique'
  return std::make_unique<ConvertTorchToTcp>({});
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ 
/usr/lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/unique_ptr.h:961:5: note: candidate template ignored: substitution failure [with _Tp = mlir::tcp::(anonymous namespace)::ConvertTorchToTcp]: deduced incomplete pack <(no value)> for template parameter '_Args'
    make_unique(_Args&&... __args)                                                                                                                                                                     
    ^                                           
    ```

sjain-stanford:
@navahgar I am about to land this under the assumption that it addresses your comment about tcp.gather codegen. Please let me know if it doesn't or if you have any other feedback, and I'll be glad to take it up in a follow-on.

PR #25, opened 2024-01-10
Lower torch transposed convolution to a custom TCP op

srinathava:
As titled, lower torch transposed convolution to a custom TCP op to avoid a mis-compilation in `TorchToTosa`.

### DIFF
--- a/lib/Conversion/TorchToTcp/TcpCustomOp.cpp
+++ b/lib/Conversion/TorchToTcp/TcpCustomOp.cpp
@@ -181,8 +181,7 @@ public:
 
     auto addListOfIntAttr = [&](const std::string &name, Value value) {
       SmallVector<int64_t> valueInt;
-      if (!matchPattern(adaptor.getStride(),
-                        m_TorchListOfConstantInts(valueInt)))
+      if (!matchPattern(value, m_TorchListOfConstantInts(valueInt)))
         return rewriter.notifyMatchFailure(op, std::string("non-const") + name +
                                                    "list unsupported");
       attrs.push_back(
--- a/lib/Conversion/TorchToTcp/TcpCustomOp.cpp
+++ b/lib/Conversion/TorchToTcp/TcpCustomOp.cpp
@@ -163,49 +163,18 @@ public:
     SmallVector<Value> operands;
     operands.push_back(adaptor.getInput());
     operands.push_back(adaptor.getWeight());
+    operands.push_back(adaptor.getBias());
+    operands.push_back(adaptor.getStride());
+    operands.push_back(adaptor.getPadding());
+    operands.push_back(adaptor.getDilation());
+    operands.push_back(adaptor.getTransposed());
+    operands.push_back(adaptor.getOutputPadding());
+    operands.push_back(adaptor.getGroups());
 
-    SmallVector<int64_t, 2> stride;
-    if (!matchPattern(adaptor.getStride(), m_TorchListOfConstantInts(stride)))
-      return rewriter.notifyMatchFailure(op,
-                                         "non-const stride list unsupported");
-    SmallVector<NamedAttribute> attrs;
-    attrs.push_back(
-        rewriter.getNamedAttr("stride", rewriter.getIndexArrayAttr(stride)));
+    auto replOp =
+        rewriter.replaceOpWithNewOp<tcp::CustomOp>(op, resultTypes, operands);
 
-    SmallVector<int64_t, 2> padding_2d;
-    if (!matchPattern(adaptor.getPadding(),
-                      m_TorchListOfConstantInts(padding_2d)))
-      return rewriter.notifyMatchFailure(op,
-                                         "non-const padding list unsupported");
-    attrs.push_back(rewriter.getNamedAttr(
-        "padding", rewriter.getIndexArrayAttr(padding_2d)));
-
-    SmallVector<int64_t, 2> dilation;
-    if (!matchPattern(adaptor.getDilation(),
-                      m_TorchListOfConstantInts(dilation)))
-      return rewriter.notifyMatchFailure(op,
-                                         "non-const dilation list unsupported");
-    attrs.push_back(rewriter.getNamedAttr(
-        "dilation", rewriter.getIndexArrayAttr(dilation)));
-
-    int64_t groups;
-    if (!matchPattern(op.getGroups(), m_TorchConstantInt(&groups)))
-      return rewriter.notifyMatchFailure(op,
-                                         "non-const group size unsupported");
-    attrs.push_back(
-        rewriter.getNamedAttr("groups", rewriter.getI64IntegerAttr(groups)));
-
-    bool transposed;
-    if (matchPattern(op.getTransposed(), m_TorchConstantBool(&transposed)))
-      return rewriter.notifyMatchFailure(
-          op, "non-const transposed bit unsupported");
-    attrs.push_back(
-        rewriter.getNamedAttr("transposed", rewriter.getBoolAttr(transposed)));
-
-    auto replOp = rewriter.replaceOpWithNewOp<tcp::CustomOp>(op, resultTypes,
-                                                             operands, attrs);
-
-    replOp.setOpName("torch.aten.convolution");
+    replOp.setOpName(op->getName().getStringRef());
 
     return success();
   }
--- a/lib/Conversion/TorchToTcp/TcpCustomOp.cpp
+++ b/lib/Conversion/TorchToTcp/TcpCustomOp.cpp
@@ -161,18 +161,68 @@ public:
     }
 
     SmallVector<Value> operands;
-    operands.push_back(adaptor.getInput());
-    operands.push_back(adaptor.getWeight());
-    operands.push_back(adaptor.getBias());
-    operands.push_back(adaptor.getStride());
-    operands.push_back(adaptor.getPadding());
-    operands.push_back(adaptor.getDilation());
-    operands.push_back(adaptor.getTransposed());
-    operands.push_back(adaptor.getOutputPadding());
-    operands.push_back(adaptor.getGroups());
+    SmallVector<StringRef> operandNames;
 
-    auto replOp =
-        rewriter.replaceOpWithNewOp<tcp::CustomOp>(op, resultTypes, operands);
+    auto addOperand = [&](std::string name, Value value) {
+      operandNames.push_back(name);
+      operands.push_back(value);
+    };
+
+    addOperand("input", adaptor.getInput());
+    addOperand("weight", adaptor.getWeight());
+    if (!adaptor.getBias().getType().isa<Torch::NoneType>()) {
+      addOperand("bias", adaptor.getBias());
+    }
+
+    SmallVector<NamedAttribute> attrs;
+
+    attrs.push_back(rewriter.getNamedAttr(
+        "torch_operand_names", rewriter.getStrArrayAttr(operandNames)));
+
+    auto addListOfIntAttr = [&](const std::string &name, Value value) {
+      SmallVector<int64_t> valueInt;
+      if (!matchPattern(adaptor.getStride(),
+                        m_TorchListOfConstantInts(valueInt)))
+        return rewriter.notifyMatchFailure(op, std::string("non-const") + name +
+                                                   "list unsupported");
+      attrs.push_back(
+          rewriter.getNamedAttr(name, rewriter.getIndexArrayAttr(valueInt)));
+      return success();
+    };
+
+    if (auto result = addListOfIntAttr("stride", adaptor.getStride());
+        result.failed()) {
+      return result;
+    }
+    if (auto result = addListOfIntAttr("padding", adaptor.getPadding());
+        result.failed()) {
+      return result;
+    }
+    if (auto result = addListOfIntAttr("dilation", adaptor.getDilation());
+        result.failed()) {
+      return result;
+    }
+    if (auto result =
+            addListOfIntAttr("output_padding", adaptor.getOutputPadding());
+        result.failed()) {
+      return result;
+    }
+
+    bool transposed;
+    if (!matchPattern(op.getTransposed(), m_TorchConstantBool(&transposed)))
+      return rewriter.notifyMatchFailure(op,
+                                         "non const transposed unsupported");
+    attrs.push_back(
+        rewriter.getNamedAttr("transposed", rewriter.getBoolAttr(transposed)));
+
+    int64_t groups;
+    if (!matchPattern(op.getGroups(), m_TorchConstantInt(&groups)))
+      return rewriter.notifyMatchFailure(op, "non const groups unsupported");
+    attrs.push_back(
+        rewriter.getNamedAttr("groups", rewriter.getI64IntegerAttr(groups)));
+
+    auto replOp = rewriter.replaceOpWithNewOp<tcp::CustomOp>(op, resultTypes,
+                                                             operands, attrs);
 
     replOp.setOpName(op->getName().getStringRef());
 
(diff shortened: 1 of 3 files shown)


lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -146,6 +147,70 @@ class ConvertAten_IndexPutImplOp
...
+                  ConversionPatternRewriter &rewriter) const override {
+    SmallVector<Type> resultTypes;
+    if (failed(OpConversionPattern<AtenConvolutionOp>::getTypeConverter()
+                   ->convertTypes(op->getResultTypes(), resultTypes))) {
+      return failure();
+    }
+
+    SmallVector<Value> operands;
+    operands.push_back(adaptor.getInput());
+    operands.push_back(adaptor.getWeight());
```
sjain-stanford, line 165: Why is the `$bias` [operand](https://sourcegraph.com/github.com/llvm/torch-mlir@4dd17f0b711523223e41606d9b7b6023b9149d46/-/blob/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td?L5578) (optional torch tensor) being dropped?

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -146,6 +147,70 @@ class ConvertAten_IndexPutImplOp
...
+    if (matchPattern(op.getTransposed(), m_TorchConstantBool(&transposed)))
+      return rewriter.notifyMatchFailure(
+          op, "non-const transposed bit unsupported");
+    attrs.push_back(
+        rewriter.getNamedAttr("transposed", rewriter.getBoolAttr(transposed)));
+
+    auto replOp = rewriter.replaceOpWithNewOp<tcp::CustomOp>(op, resultTypes,
+                                                             operands, attrs);
+
+    replOp.setOpName("torch.aten.convolution");
```
sjain-stanford, line 208: nit: `replOp.setOpName(op->getName().getStringRef());` would be cleaner

test/Conversion/TorchToTcp/tcp_custom_ops.mlir
```diff
@@ -57,3 +57,18 @@ func.func @torch.aten.index_put_impl_op(%arg0: !torch.vtensor<[25],f32>, %arg1:
   %1 = torch.aten._index_put_impl %arg0, %0, %arg2, %false, %false : !torch.vtensor<[25],f32>, !torch.list<optional<vtensor>>, !torch.vtensor<[],f32>, !torch.bool, !torch.bool -> !torch.vtensor<[25],f32>
   return %1 : !torch.vtensor<[25],f32>
 }
+
+// ---
+
+// CHECK: tcp.custom_op("tcp.torch_transposed_conv2d") %{{.*}}, %{{.*}} {dilation = [1 : index, 1 : index], groups = 1 : i64, padding = [1 : index, 1 : index], stride = [2 : index, 2 : index]} : tensor<1x64x1x100xf32>, tensor<64x64x3x3xf32> -> tensor<1x64x2x200xf32>
```
sjain-stanford, line 63: Fix op name here? 
`s/tcp.torch_transposed_conv2d/torch.aten.convolution/`

lib/Conversion/TorchToTcp/Utils.h
```diff
@@ -80,7 +80,10 @@ bool getConstTensorWithType(ConversionPatternRewriter &rewriter, Operation *op,
 template <typename TorchToTcpPattern, typename AtenOp>
 inline void addPatternIfOpInConvertTorchOpsSet(
     TypeConverter &typeConverter, RewritePatternSet &patterns,
-    ConversionTarget &target, const llvm::StringSet<> &convertTorchOpsSet) {
+    ConversionTarget &target, const llvm::StringSet<> &convertTorchOpsSet,
+    std::function<bool(AtenOp)> dynamicLegalityFcn = [](AtenOp) {
```
sjain-stanford, line 84: Looks like the `dynamicLegalityFcn` is not being used below? I think you want this to be used to mark the op dynamically legal?

test/Conversion/TorchToTcp/tcp_custom_ops.mlir
```diff
@@ -57,3 +57,18 @@ func.func @torch.aten.index_put_impl_op(%arg0: !torch.vtensor<[25],f32>, %arg1:
...
+func.func @forward(%input: !torch.vtensor<[1,64,1,100],f32>) -> !torch.vtensor<[1,64,2,200],f32> {
+  %true = torch.constant.bool true
+  %int1 = torch.constant.int 1
+  %int2 = torch.constant.int 2
+  %weight = torch.vtensor.literal(dense<0.0> : tensor<64x64x3x3xf32>) : !torch.vtensor<[64,64,3,3],f32>
+  %bias = torch.vtensor.literal(dense<0.0> : tensor<64xf32>) : !torch.vtensor<[64],f32>
+  %stride = torch.prim.ListConstruct %int2, %int2 : (!torch.int, !torch.int) -> !torch.list<int>
+  %int1x1 = torch.prim.ListConstruct %int1, %int1 : (!torch.int, !torch.int) -> !torch.list<int>
+  %output = torch.aten.convolution %input, %weight, %bias, %stride, %int1x1, %int1x1, %true, %int1x1, %int1 : !torch.vtensor<[1,64,1,100],f32>, !torch.vtensor<[64,64,3,3],f32>, !torch.vtensor<[64],f32>, !torch.list<int>, !torch.list<int>, !torch.list<int>, !torch.bool, !torch.list<int>, !torch.int -> !torch.vtensor<[1,64,2,200],f32>
```
sjain-stanford, line 72: Consider adding a negative test for regular convolution (i.e. torch.aten.convolution with transposed `false`) where the conversion to tcp.custom_op doesn't take place.

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -159,4 +224,15 @@ void torch_to_tcp::populateTcpCustomOpPatternsAndLegality(
   INSERT_ATEN_TO_TCP_CUSTOM_OP_PATTERN(AtenIndexTensorHackedTwinOp);
   INSERT_ATEN_TO_TCP_CUSTOM_OP_PATTERN(Aten_IndexPutImplOp);
 #undef INSERT_ATEN_TO_TCP_CUSTOM_OP_PATTERN
+
+  auto isNonTransposedConvOp = [](AtenConvolutionOp op) {
+    bool transposed = false;
+    if (matchPattern(op.getTransposed(), m_TorchConstantBool(&transposed)))
+      return !transposed;
+    return false;
+  };
```
sjain-stanford, line 233: (minor nit, feel free to push back) It is a bit difficult to interpret from this, so asking here: 
What is the convention for `DynamicLegalityCallbackFn`? In other words can we frame this as `isTransposedConvOp` instead for readability?

srinathava: It is going to be passed to `addDynamicallyLegalOp` which is why we need to define when an op is legal instead of needing to be lowered.

I agree that this means that we have to use a "negated" function because often what we want to specify is what we want to lower instead.

I wish MLIR allowed `addDynamicallyIllegalOp` but I do not see it. I vote to keep it as is for now.

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -146,6 +147,70 @@ class ConvertAten_IndexPutImplOp
...
+
+    SmallVector<int64_t, 2> stride;
+    if (!matchPattern(adaptor.getStride(), m_TorchListOfConstantInts(stride)))
+      return rewriter.notifyMatchFailure(op,
+                                         "non-const stride list unsupported");
+    SmallVector<NamedAttribute> attrs;
+    attrs.push_back(
+        rewriter.getNamedAttr("stride", rewriter.getIndexArrayAttr(stride)));
+
+    SmallVector<int64_t, 2> padding_2d;
```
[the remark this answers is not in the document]
srinathava: Given all the complexity inherent in doing a pattern match here (without actually exploiting it beyond just creating a named attr), I have migrated this op to just transfer all the original operands into operands of the `tcp.custom_op`. While this means that the eventual lowering of this to a kernel will have to use indices instead of names to recover them, it avoids this kind of hard-coding and brittleness.

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -146,6 +147,39 @@ class ConvertAten_IndexPutImplOp
...
+  matchAndRewrite(AtenConvolutionOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    SmallVector<Type> resultTypes;
+    if (failed(OpConversionPattern<AtenConvolutionOp>::getTypeConverter()
+                   ->convertTypes(op->getResultTypes(), resultTypes))) {
+      return failure();
+    }
+
+    SmallVector<Value> operands;
+    operands.push_back(adaptor.getInput());
```
srinathava, line 164: Actually, I wonder if there's just a generic iterator over all the input parameters of the original op so I do not have to write this down manually. Will get back to this in a bit.

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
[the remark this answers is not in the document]
sjain-stanford: > Given all the complexity inherent in doing a pattern match here 
Can we use `SmallVector<Value>` here instead? I didn't mean to say we should use a `SmallVector` of the exact size (depending on 1d, 2d, 3d etc), but leave it generic.

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -146,6 +147,89 @@ class ConvertAten_IndexPutImplOp
...
+      return failure();
+    }
+
+    SmallVector<Value> operands;
+    SmallVector<StringRef> operandNames;
+
+    auto addOperand = [&](std::string name, Value value) {
+      operandNames.push_back(name);
+      operands.push_back(value);
+    };
```
sjain-stanford, line 169: Non-blocking and could be addressed in a follow-on PR:
This would be nice to factor out as a utility, so they can be reused by other TcpCustomOp conversion patterns.

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -146,6 +147,89 @@ class ConvertAten_IndexPutImplOp
...
+    auto addListOfIntAttr = [&](const std::string &name, Value value) {
+      SmallVector<int64_t> valueInt;
+      if (!matchPattern(adaptor.getStride(),
+                        m_TorchListOfConstantInts(valueInt)))
+        return rewriter.notifyMatchFailure(op, std::string("non-const") + name +
+                                                   "list unsupported");
+      attrs.push_back(
+          rewriter.getNamedAttr(name, rewriter.getIndexArrayAttr(valueInt)));
+      return success();
+    };
```
sjain-stanford, line 191: same

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -146,6 +147,89 @@ class ConvertAten_IndexPutImplOp
...
+  matchAndRewrite(AtenConvolutionOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    SmallVector<Type> resultTypes;
+    if (failed(OpConversionPattern<AtenConvolutionOp>::getTypeConverter()
+                   ->convertTypes(op->getResultTypes(), resultTypes))) {
+      return failure();
+    }
+
+    SmallVector<Value> operands;
+    SmallVector<StringRef> operandNames;
```
sjain-stanford, line 164: Naming nit: Prefer `tensorOperandNames` (to distinguish from non-tensor operands that we convert to attributes).

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -146,6 +147,89 @@ class ConvertAten_IndexPutImplOp
...
+
+    SmallVector<NamedAttribute> attrs;
+
+    attrs.push_back(rewriter.getNamedAttr(
+        "torch_operand_names", rewriter.getStrArrayAttr(operandNames)));
+
+    auto addListOfIntAttr = [&](const std::string &name, Value value) {
+      SmallVector<int64_t> valueInt;
+      if (!matchPattern(adaptor.getStride(),
+                        m_TorchListOfConstantInts(valueInt)))
```
sjain-stanford, line 185: I think this has a bug. Should this be
```
if (!matchPattern(value, m_TorchListOfConstantInts(valueInt)))
```
instead?

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
sjain-stanford: actually, never-mind since we seem to use operands to refer to tensorOperands always so there is no ambiguity..

test/Conversion/TorchToTcp/tcp_custom_ops.mlir
```diff
@@ -57,3 +57,28 @@ func.func @torch.aten.index_put_impl_op(%arg0: !torch.vtensor<[25],f32>, %arg1:
...
+
+  %input2 = torch.vtensor.literal(dense<0.0> : tensor<1x9x16x1600xf32>) : !torch.vtensor<[1,9,16,1600],f32>
+  %weights2 = torch.vtensor.literal(dense<0.0> : tensor<32x9x3x3xf32>) : !torch.vtensor<[32,9,3,3],f32>
+  %int0x0 = torch.prim.ListConstruct %int0, %int0 : (!torch.int, !torch.int) -> !torch.list<int>
+  %none = torch.constant.none
+  %output2 = torch.aten.convolution %input2, %weights2, %none, %int1x1, %int1x1, %int1x1, %false, %int0x0, %int1 : !torch.vtensor<[1,9,16,1600],f32>, !torch.vtensor<[32,9,3,3],f32>, !torch.none, !torch.list<int>, !torch.list<int>, !torch.list<int>, !torch.bool, !torch.list<int>, !torch.int -> !torch.vtensor<[1,32,16,1600],f32>
```
sjain-stanford, line 81: Please consider splitting this into a separate test (in the same file) for readability and/or ease of maintenance.

test/Conversion/TorchToTcp/tcp_custom_ops.mlir
```diff
@@ -57,3 +57,28 @@ func.func @torch.aten.index_put_impl_op(%arg0: !torch.vtensor<[25],f32>, %arg1:
   %1 = torch.aten._index_put_impl %arg0, %0, %arg2, %false, %false : !torch.vtensor<[25],f32>, !torch.list<optional<vtensor>>, !torch.vtensor<[],f32>, !torch.bool, !torch.bool -> !torch.vtensor<[25],f32>
   return %1 : !torch.vtensor<[25],f32>
 }
+
+// ---
+
+// CHECK: tcp.custom_op("torch.aten.convolution") %{{.*}}, %{{.*}}, %{{.*}} {dilation = [2 : index, 2 : index], groups = 1 : i64, output_padding = [2 : index, 2 : index], padding = [2 : index, 2 : index], stride = [2 : index, 2 : index], torch_operand_names = ["input", "weight", "bias"], transposed = true} : tensor<1x64x1x100xf32>, tensor<64x64x3x3xf32>, tensor<64xf32> -> tensor<1x64x2x200xf32>
```
sjain-stanford, line 63: I was wondering how the test passes despite the bug above. It seems to be because padding is being populated from stride and the check statement propagates the bug. 
```
// This should've been padding = [1 : index, 1 : index]
padding = [2 : index, 2 : index]
```

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
srinathava: Yikes :( Fixed.

test/Conversion/TorchToTcp/tcp_custom_ops.mlir
srinathava: good suggestion. I split it and also split the CHECK line into multiple lines to make it a bit clearer.

test/Conversion/TorchToTcp/tcp_custom_ops.mlir
srinathava: :( This PR is turning out to be a bit of typo central. Thanks for the careful review.

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -19,6 +19,7 @@
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 
 #include "llvm/ADT/StringSet.h"
+#include "llvm/Support/Debug.h"
```
sjain-stanford, line 22: Just checking, is this still needed?

sjain-stanford (approved):
Looks great to me. Thanks for contributing and persevering through the iterations!

PR #28, opened 2024-01-16
[TorchToTcp] Make tensor operand checks in `aten.convolution -> tcp.custom_op` lit test explicit

sjain-stanford:
As titled.


srinathava (approved):
LGTM.

PR #30, opened 2024-01-17
Support clangd for code completion, navigation and insights

sjain-stanford:
Uses [bazel-compile-commands-extractor](https://github.com/hedronvision/bazel-compile-commands-extractor) to generate the compilation database from our bazel build. 

Other minor improvements include:
- simplify .bazelrc
- move buildifier and compile-commands-extractor targets to separate `tools/` to make downstream integrations easier (else they fail on loading these deps)
- bump bazel to 6.4.0 (for `--noenable_bzlmod` option)


navahgar (approved):
LGTM

PR #32, opened 2024-01-18
[Bazel] Auto detect platform specific configs

sjain-stanford:
This let's us drop the annoying `--config=clang_linux` from every bazel build/run/test command. Also simplifies clangd compilation database generation PR a bit (follow-on).


navahgar (approved):
Awesome. LGTM.

PR #33, opened 2024-01-18
Add TCP custom op builder helper

srinathava:
Add a utility to aid in converting torch ops to `tcp.custom_op`


lib/Conversion/TorchToTcp/TcpCustomOpBuilder.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include "mlir-tcp/Conversion/TorchToTcp/TcpCustomOpBuilder.h"
+
+#include "torch-mlir/Conversion/Utils/Utils.h"
+#include "torch-mlir/Dialect/Torch/Utils/Utils.h"
+
+namespace mlir::tcp {
+
+void TcpCustomOpBuilder::addOperand(std::string opName, Value value) {
+  // Should there be some error checking here that Value has a specific
+  // type (for instance a RankedTensorType?)
```
navahgar, line 19: It should conform to Tcp_Tensor which is defined here: https://github.com/cruise-automation/mlir-tcp/blob/main/include/mlir-tcp/Dialect/IR/TcpTypes.td#L54. But that should automatically happen with the verifier. So, there is no need to explicit error checking here.

include/mlir-tcp/Conversion/TorchToTcp/TcpCustomOpBuilder.h
```diff
@@ -0,0 +1,71 @@
...
+  void addAsMultipleTensorOperands(std::string opNamePrefix, mlir::Value value);
+
+  // Add value as a named bool attribute
+  void addBoolAttr(std::string attrName, Value value);
+
+  // Add value as a named integer attribute
+  void addIntAttr(std::string attrName, Value value);
+
+  // Add value as a named list of integers attribute
+  void addListOfIntsAttr(std::string attrName, Value value);
```
navahgar, line 56: Do you have any strong reasons to have these as methods on a class as compared to stand-alone functions?

Here are the advantages I see with stand-alone functions:
* These functions can be used through this conversion code (not just the custom op conversions) without instantiating any objects. We don't have to keep track of the object lifetime in that case.
* Builder is already over-loaded in MLIR. Adding one more here seems confusing.
* That is the approach followed in MLIR AFAICT. So, there will be uniformity.

So, I strongly recommend moving to stand-alone functions.

navahgar: @sjain-stanford Can you think of other reasons for going with stand-alone functions?

navahgar (commented):
Thanks for taking this up.

include/mlir-tcp/Conversion/TorchToTcp/TcpCustomOpBuilder.h
srinathava: The reason I went with methods is that as we call them, we are accumulating the state of the built-up tcp.custom_op's operands etc. into `operandNames`, `operands`, `conversionResult`, `attrs`. Without these things being methods on some class, all of that state will have to be hoisted into the callee function and passed in. At that point, the code basically becomes "classes in C" as opposed to a C++ class. At least I cannot immediately imagine writing the callee-side code as cleanly without this encapsulation. See for instance, this bit:

```cpp
    builder.addOperand("self", adaptor.getSelf());
    builder.addOperand("index", adaptor.getIndex());
    builder.addIntAttr("axis", op.getDim());

    return builder.replace();
```

Without the class, would we have to declare the members of the class in that function (operandNames etc.) and then pass these in by reference to these (now free) functions? How would that be really different?

include/mlir-tcp/Conversion/TorchToTcp/TorchToTcpCustomOpConversionHelper.h
```diff
@@ -0,0 +1,72 @@
...
+///
+/// 1. Some of the tensor operands to the original torch op are converted
+/// into positional tensor operands. Given that a tcp.custom_op does not
+/// have a strongly typed API to fetch the operands other than by index, we
+/// also add an extra StringArray attribute to the tcp.custom_op called
+/// "torch_operand_names" which describes what each positional argument of
+/// the `tcp.custom_op` corresponds to. This also allows us to handle the
+/// case of optional tensor arguments.
+///
+/// 2. Some of the tensor operands are converted into named attributes of
```
sjain-stanford, line 30: I believe this should say `non-tensor operands`, since we keep all tensor operands as-is.

BUILD
```diff
@@ -195,12 +195,14 @@ cc_library(
         "lib/Conversion/TorchToTcp/TcpCustomOp.cpp",
         "lib/Conversion/TorchToTcp/TorchToTcp.cpp",
         "lib/Conversion/TorchToTcp/TorchToTcpCustomOp.cpp",
+        "lib/Conversion/TorchToTcp/TorchToTcpCustomOpConversionHelper.cpp",
```
sjain-stanford, line 198: Prefer using the existing `lib/Conversion/TorchToTcp/Utils.cpp` instead?

lib/Conversion/TorchToTcp/TorchToTcpCustomOpConversionHelper.cpp
```diff
@@ -0,0 +1,116 @@
...
+  auto result = typeConverter->convertTypes(op->getResultTypes(), resultTypes);
+  if (result.failed()) {
+    return result;
+  }
+
+  // TODO: Is there some better way to convert a
+  // SmallVector<std::string> to SmallVector<StringRef> locally here? Note
+  // that we cannot store operandNames directly as a SmallVector<StringRef>
+  // because the strings passed to addOperand etc. could be computed
+  // temporary objects.
```
sjain-stanford, line 57: remove TODO as we last discussed

BUILD
navahgar: One alternative is to create a `TorchToTcpCustomOpUtils.h` and `.cpp`, to hold those that are specific to custom op cases. Wdyt?

navahgar (commented):
LGTM. I'll let @sjain-stanford approve once his comment is addressed.

BUILD
sjain-stanford: Given that `TorchToTcpCustomOp` lib  is not factored as a separate conversion target (i.e. it reuses `TorchToTcp` [target](https://github.com/cruise-automation/mlir-tcp/blob/main/BUILD#L188)), and we already have some utilities in `Utils.cpp` that are shared by both (e.g. `addPatternIfOpInConvertTorchOpsSet`), it might make sense to keep them colocated. There are organizational benefits in doing so, like making future refactorings for code reuse easier.

lib/Conversion/TorchToTcp/TorchToTcpCustomOpConversionHelper.cpp
```diff
@@ -0,0 +1,116 @@
...
+                                                    Value value) {
+  // Should there be some error checking here that Value has a specific
+  // type (for instance a RankedTensorType?)
+  operandNames.push_back(opName);
+  operands.push_back(value);
+}
+
+void TorchToTcpCustomOpConversionHelper::addAsMultipleTensorOperands(
+    std::string opNamePrefix, mlir::Value value) {
+  mlir::SmallVector<Value> indicesTorchType;
```
sjain-stanford, line 27: Do we need the early return here too (like in the other methods)?
```
  if (conversionResult.failed())
    return;
```

lib/Conversion/TorchToTcp/TorchToTcpCustomOpConversionHelper.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include "mlir-tcp/Conversion/TorchToTcp/TorchToTcpCustomOpConversionHelper.h"
+
+#include "torch-mlir/Conversion/Utils/Utils.h"
+#include "torch-mlir/Dialect/Torch/Utils/Utils.h"
+
+namespace mlir::tcp {
+
+void TorchToTcpCustomOpConversionHelper::addOperand(std::string opName,
+                                                    Value value) {
+  // Should there be some error checking here that Value has a specific
```
sjain-stanford, line 19: Do we need the early return here too (like in the other methods)?
```
  if (conversionResult.failed())
    return;
```

include/mlir-tcp/Conversion/TorchToTcp/TorchToTcpCustomOpConversionHelper.h
srinathava: I'll change "tensor operands" to just "operands". My understanding is that a torch OP _can_ have a tensor operand which we merely insist is a constant.

BUILD
srinathava: Interesting... In a previous life, we preferred NOT having files like "utils" etc. because they would become a general catch-all of eventually unrelated things and preferred more standalone headers (ideally a header per class decl.) to make dependencies in the C++ file much more granular.

I have no strong preferences here. @sjain-stanford , did you mean to just put the implementation (cpp file) into `Utils.cpp` or also the helper class into `Utils.h`?

sjain-stanford: I'm going by patterns from the `llvm-project`/`torch-mlir` codebases (which we try to mimic in TCP). See some examples below:
- https://github.com/llvm/torch-mlir/blob/main/lib/Conversion/Utils/Utils.cpp
- https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/SCF/Utils/Utils.cpp
- https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/Tosa/Utils/ConversionUtils.cpp

> did you mean to just put the implementation (cpp file) into Utils.cpp or also the helper class into Utils.h?
Let's move the class interface to `Utils.h` as well (this is already included for use by the conversion patterns).

include/mlir-tcp/Conversion/TorchToTcp/TorchToTcpCustomOpConversionHelper.h
sjain-stanford: >  My understanding is that a torch OP can have a tensor operand which we merely insist is a constant.

FWICT we always preserve tensor operands as tensors. Yes, scalar types get converted to attributes, sometimes even if they're not constants. @navahgar Can you think of a case otherwise?

[TorchTypes.td](https://sourcegraph.com/github.com/llvm/torch-mlir@main/-/blob/include/torch-mlir/Dialect/Torch/IR/TorchTypes.td) has a clear separation between tensor and non-tensor (aka scalar) dtypes. The regular types like `Torch_IntType` or `Torch_FloatType` are explicitly modeled after python ints/floats and it notes `This type is not used for modeling tensor dtypes`. 
```
def Torch_IntType : Torch_Type<"Int", "int"> {
  let summary = "Torch IntType";
  let description = [{
    The integer type used to model the Python `int` type in TorchScript.
    TorchScript itself models this type as a 64-bit signed integer.

    Note: This type is not used for modeling tensor dtypes.
  }];
}
```
Please correct me if I'm missing something.

lib/Conversion/TorchToTcp/Utils.cpp
```diff
@@ -389,3 +382,101 @@ Value torch_to_tcp::scalarToTcpTensor(ConversionPatternRewriter &rewriter,
...
+        "unimplemented: the tensor list is not from list construct");
+    return;
+  }
+
+  mlir::SmallVector<Value> indexTensors = torch::Torch::getTypeConvertedValues(
+      rewriter, op->getLoc(), typeConverter, indicesTorchType);
+
+  for (size_t i = 0; i < indexTensors.size(); ++i) {
+    addOperand(opNamePrefix + std::to_string(i), indexTensors[i]);
+  }
```
sjain-stanford, line 410: Can we make this conditional on `indexTensors.size() > 1` (to avoid the single element list case having a redundant `_0` suffix)

sjain-stanford (approved):
LGTM. Thanks for the thoughtful refactoring - this significantly simplifies writing and maintaining TCP custom op conversions.

include/mlir-tcp/Conversion/TorchToTcp/TorchToTcpCustomOpConversionHelper.h
navahgar: > > My understanding is that a torch OP can have a tensor operand which we merely insist is a constant.
> 
> FWICT we always preserve tensor operands as tensors. 

That's right. Tensor operands are always inputs to the ops. At this point during the lowering, I can't think of a case where we insist this to be constants.

But later in the compilation, we have some cases, like in the QuantSim ops, where we insist that they are constants.

> Yes, scalar types get converted to attributes, sometimes even if they're not constants.

One clarification here: they have to be constants to be converted to attributes. We probably error out if they are not constants in such cases.

PR #35, opened 2024-01-22
[TorchToTcp] lowering support for aten.size.int op

zezhang:
Lowering `aten.size.int` op to `tensor::dim` op during torch-to-tcp.

Test (in docker):

`bazel test //test/...`

### DIFF
--- a/lib/Conversion/TorchToTcp/Misc.cpp
+++ b/lib/Conversion/TorchToTcp/Misc.cpp
@@ -133,6 +133,38 @@ public:
   }
 };
 
+class ConvertAtenSizeIntOp
+    : public OpConversionPattern<AtenSizeIntOp> {
+public:
+  using OpConversionPattern<AtenSizeIntOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(AtenSizeIntOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op->getLoc();
+    Value self = adaptor.getSelf();
+    auto type = self.getType().cast<RankedTensorType>();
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchToTcp/Misc.cpp
```diff
@@ -133,6 +133,39 @@ class ConvertValueTensorLiteralOp
...
+    }
+    auto idxOp = rewriter.create<arith::ConstantIndexOp>(loc, idxVal);
+    auto dimOp = rewriter.create<tensor::DimOp>(loc, self, idxOp);
+    auto result =
+        rewriter.create<arith::IndexCastOp>(loc, rewriter.getI64Type(), dimOp);
+
+    rewriter.replaceOp(op, result);
+    if (constIntOp->hasOneUse()) {
+      rewriter.eraseOp(constIntOp);
+    }
```
navahgar, line 163: Why is this necessary? Shouldn't this be automatically taken care of by DCE pass (canonicalizer)?

test/Conversion/TorchToTcp/misc.mlir
navahgar, line 1: I'm assuming most the changes in this file (except the new test below) are just reformatting the existing tests.

lib/Conversion/TorchToTcp/Misc.cpp
zezhang: Good point. Just removed.

test/Conversion/TorchToTcp/misc.mlir
zezhang: Yes, that's correct. Only added a new test for the aten.size.int case. Rest just fix the formatting.

navahgar (approved):
LGTM

PR #37, opened 2024-02-04
Hermetic python, TorchDynamo/FX frontend and TorchToTcp conversion for dynamo exported models

sjain-stanford:
This PR includes support for calling into the TorchDynamo export + FX import APIs in torch-mlir to generate `Torch` dialect, followed by conversions to TCP dialect. This is bundled together with a fully functional hermetic python sandbox in TCP bazel, with access to torch-mlir and core mlir's python APIs and ability to run python lit/integration style tests.

This depended on a bunch of foundational work:
1. Add a hermetic python interpreter to `mlir-tcp` bazel: Needed for mlir/torch-mlir pybind integrations, python lit+integration tests.
2. Register python bindings for torch-mlir/core mlir: Upstreaming torch-mlir's bazel python build had several issues (as it's heavily dependent on Cruise's internal bazel workspace), so I decided to instead install from pre-built python wheels (for torch and torch-mlir). This allows keeping the setup in TCP lean, and leverage/reuse torch-mlir's Cmake python build for any python integrations.
3. Automate releases of torch-mlir python binaries (https://github.com/llvm/torch-mlir-release/pull/1): This was [discontinued](https://discourse.llvm.org/t/rfc-discontinuing-pytorch-1-binary-releases/76371) as a part of the CI revamp.
4. Support for running python lit tests: Updated lit configs with python interpreter integration.
5. Add `fx_importer/basic_test.py` (as a python lit test) and mlir lit tests for TorchToTcp conversions.

Other minor workflow updates:
- Use `--test_output=errors` for better log readability in CI failures
- Add setup-build/action.yml to consolidate workspace setup
- Include disk space cleanup actions (due to OOM issues with default GHA runners)
- Bump actions versions

P.S. Our stablehlo build fails at the moment due to disk space issues (unrelated to this change, verified it passes locally). This is just an outcome of using default GitHub hosted runners for CI with very limited disk space resources. Adding some cleanup steps helped with llvm build but still fails stablehlo. We might want to look into self-hosted runners with larger memory/compute going forward. For now the non-TCP workflows are not merge-gating (optional) so this is not a blocker to this PR landing.


test/Conversion/TcpToArith/basic.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: tcp-opt %s -convert-tcp-to-arith -split-input-file | FileCheck %s
+// RUN: tcp-opt %s -convert-tcp-to-arith | FileCheck %s
```
navahgar, line 1: minor: Why remove `-split-input-file`? Are you expecting this file to not have more than 1 test?

tools/aot/BUILD
```diff
@@ -1,3 +1,8 @@
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# Also available under a BSD-style license. See LICENSE.
+
```
navahgar, line 5: Is this license common for repos in `cruise-automation`?  Do we need to get any permission?

navahgar (approved):
LGTM. Thanks for doing this.

Minor comments inline.

AaronStGeorge (approved):
Looks good to me

test/Conversion/TcpToArith/basic.mlir
sjain-stanford: Yeah just a cleanup for consistency. Let's add this back when we have more tests per file.

tools/aot/BUILD
sjain-stanford: This is the legal approved header and in use in most of our source files. Adding it here as it was missed initially.

PR #38, opened 2024-02-12
Disable third party builds (on push to main) until we migrate to self-hosted runners

sjain-stanford:
We're hitting `no space left on device` errors during llvm and stablehlo builds with the free GitHub hosted runners, which come with limited disk space allocations. Cleaning up the runners as recommended [here](https://github.com/orgs/community/discussions/25678), [here](https://github.com/orgs/community/discussions/26351) and https://github.com/actions/runner-images/issues/2875 have helped partially (increase root mount from 20G to 47G) but it still isn't enough for some builds. Temporarily disable the 3p builds on push to main (they will still run as non-merge gating on PRs), and re-enable once self-hosted runners are setup.


navahgar (approved):
LGTM

PR #39, opened 2024-02-16
Lower aten.slice.Tensor to tensor.extract_slice

zezhang:
As titled.

Test (in docker):

`bazel test //test/...`


lib/Conversion/TorchToTcp/DataMovement.cpp
```diff
@@ -26,6 +28,75 @@ using namespace mlir::torch;
 using namespace mlir::torch::Torch;
 
 namespace {
+template <typename OpTy, typename OpAdaptor>
```
sjain-stanford, line 31: Do you expect this to be useful for ops other than `AtenSliceTensorOp`? Just wondering if there's a need to templatize.

test/Conversion/TorchToTcp/data_movement.mlir
```diff
@@ -11,3 +11,17 @@ func.func @torch.aten.cat(%arg0: !torch.vtensor<[?,?],f32>, %arg1: !torch.vtenso
...
+//   CHECK-SAME:   %[[ARG0:.+]]: !torch.vtensor<[1,56,?,?],f32>
+//        CHECK:   %[[V1:.+]] = torch_c.to_builtin_tensor %[[ARG0]] : !torch.vtensor<[1,56,?,?],f32> -> tensor<1x56x?x?xf32>
+//        CHECK:   %[[V2:.+]] = tensor.extract_slice %[[V1]][%{{.*}}, %{{.*}}, %{{.*}}, %{{.*}}] [%{{.*}}, %{{.*}}, %{{.*}}, %{{.*}}] [%{{.*}}, %{{.*}}, %{{.*}}, %{{.*}}] : tensor<1x56x?x?xf32> to tensor<?x?x?x?xf32>
+//        CHECK:   %[[V3:.+]] = tensor.cast %[[V2]] : tensor<?x?x?x?xf32> to tensor<1x28x?x?xf32>
+func.func @torch.aten.slice.Tensor(%arg0: !torch.vtensor<[1,56,?,?],f32>) -> !torch.vtensor<[1,28,?,?],f32> {
+  %int100 = torch.constant.int 100
+  %int1 = torch.constant.int 1
+  %int0 = torch.constant.int 0
+  %int2 = torch.constant.int 2
+  %0 = torch.aten.slice.Tensor %arg0, %int1, %int0, %int100, %int2 : !torch.vtensor<[1,56,?,?],f32>, !torch.int, !torch.int, !torch.int, !torch.int -> !torch.vtensor<[1,28,?,?],f32>
```
sjain-stanford, line 25: Going by the [docs](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorextract_slice-tensorextractsliceop) for `tensor.extract_slice`, a reduction in rank is allowed. Does `aten.slice.Tensor` operate the same way, and if not, do we need to add an extra check to ensure we're lowering this consistent with the semantics of `aten.slice.Tensor`?

test/Conversion/TorchToTcp/data_movement.mlir
```diff
@@ -11,3 +11,17 @@ func.func @torch.aten.cat(%arg0: !torch.vtensor<[?,?],f32>, %arg1: !torch.vtenso
   %1 = torch.aten.cat %0, %int0 : !torch.list<vtensor>, !torch.int -> !torch.vtensor<[?,?],f32>
   return %1 : !torch.vtensor<[?,?],f32>
 }
+
+// CHECK-LABEL: @torch.aten.slice.Tensor
```
sjain-stanford, line 15: nit: Specify `-split-input-file` and separate the tests with the `// -----` delimiter? Curious how it worked without that.

lib/Conversion/TorchToTcp/DataMovement.cpp
```diff
@@ -26,6 +28,75 @@ using namespace mlir::torch;
...
+LogicalResult prepareArgumentsForSlicingOp(OpTy op, OpAdaptor adaptor,
```
sjain-stanford, line 32: If this is a copy-paste from [here](https://sourcegraph.com/github.com/llvm/torch-mlir@d65925a8b465d4a84be947d37197178d5c5cc6d2/-/blob/lib/Conversion/TorchToLinalg/DataMovement.cpp?L42), why not call it directly instead? Was it just not being exposed (via a header)? If so, maybe say "this was adapted from ..." add a hyperlink for future reference, and specify changes you made (if any).

navahgar: +1 to this. 

We could also consider updating TorchMLIR to expose this API via some header and use it directly here, instead of copying it. Doing this in a later PR is fine. If you plan to do that, please add a TODO to stating that.

test/Conversion/TorchToTcp/data_movement.mlir
navahgar: We should confirm `aten.slice.Tensor` does not support this case. If so (which I'm assuming is the case), it seems enough to add a comment here documenting that and ensuring we don't create `tensor.extract_slice` with any mask that does rank reduction. 

Wdyt?

sjain-stanford: > We should confirm aten.slice.Tensor does not support this case.

+1. An easy way to check PyTorch semantics is using python lit tests, starting with torch.nn.Module and converting to Torch dialect using the dynamo frontend. See [example](https://github.com/cruise-automation/mlir-tcp/blob/main/test/python/fx_import/basic_test.py). @zezhang can you please add this?

If we realize aten.slice.Tensor does allow rank reduction (similar to tensor.extract_slice), please ignore the following:

> ensuring we don't create tensor.extract_slice with any mask that does rank reduction.

I'm wondering about the case when a given set of offset+size+stride values that does _not_ incur rank reduction with `aten.slice.Tensor` but does so when translated to `tensor.extract_slice`, which seems to be when there's a leading dimension with 1 in the size and stride going by this example?
```
%1 = tensor.extract_slice %0[0, 0, 0][1, 16, 4][1, 1, 1] :
  tensor<8x16x4xf32> to tensor<16x4xf32>
```
Is there an option to disable rank reduction? If not, 1) having an explicit match failure during conversion, or 2) adding an expand shape to match semantics might be better.

lib/Conversion/TorchToTcp/DataMovement.cpp
zezhang: Not for now. This part of code is copied from [here](https://sourcegraph.com/github.com/llvm/torch-mlir@d65925a8b465d4a84be947d37197178d5c5cc6d2/-/blob/lib/Conversion/TorchToLinalg/DataMovement.cpp?L42). Will add a note about it.

lib/Conversion/TorchToTcp/DataMovement.cpp
zezhang: Good point. Added.

test/Conversion/TorchToTcp/data_movement.mlir
zezhang: Good point. I have tried multiple examples, including both leading dim size of 1, and slice size of 1. Looks like the replaced `ExtractSliceOp` + `CastOp` never performs the rank reduction. This is becuase we have:

lib/Conversion/TorchToTcp/DataMovement.cpp
```diff
@@ -26,6 +28,77 @@ using namespace mlir::torch;
 using namespace mlir::torch::Torch;
 
 namespace {
+// Following function is copied from
+// https://sourcegraph.com/github.com/llvm/torch-mlir@main/-/blob/lib/Conversion/TorchToLinalg/DataMovement.cpp?L42
```
sjain-stanford, line 32: nit: use permalink here: https://sourcegraph.com/github.com/llvm/torch-mlir@4446fa00d8258311867496fc79d0b1dddd22a972/-/blob/lib/Conversion/TorchToLinalg/DataMovement.cpp?L42

test/Conversion/TorchToTcp/data_movement.mlir
```diff
@@ -1,4 +1,6 @@
-// RUN: tcp-opt %s -convert-torch-to-tcp | FileCheck %s
+// RUN: tcp-opt %s -convert-torch-to-tcp -split-input-file | FileCheck %s
+
+// -----
```
sjain-stanford, line 3: nit: the first one can be dropped

sjain-stanford (approved):
Thanks for supporting this @zezhang . LGTM.

PR #40, opened 2024-02-20
Move AotCompile to its own BUILD

sjain-stanford:
Keeps the main test/BUILD specific to LIT, and allows (as a follow-on) expanding AotCompile into e2e integration style tests for TCP based compilations.


AaronStGeorge (approved):
LGTM

PR #43, opened 2024-02-22
RefBackend E2E

sjain-stanford:


sjain-stanford:
This will require torch-mlir's python bindings to be built with tcp dialects/pipeline registrations. For now planning to use aot compile approach and build machinery around that. Closing.

PR #47, opened 2024-02-26
[AOT Compile] Use `StridedMemRefType` descriptor from upstream mlir/CRunnerUtils.h

sjain-stanford:
As titled.

### DIFF
--- a/test/AotCompile/test_aot_compiled_basic_tcp_ops.cpp
+++ b/test/AotCompile/test_aot_compiled_basic_tcp_ops.cpp
@@ -12,7 +12,6 @@
 
 #include "gtest/gtest.h"
 
-using namespace mlir;
 using namespace mlir::tcp;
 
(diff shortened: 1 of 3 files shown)


test/AotCompile/test_aot_compiled_basic_tcp_ops.cpp
```diff
@@ -7,27 +7,29 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "mlir/ExecutionEngine/CRunnerUtils.h"
 #include "tools/aot/abi.h"
 
 #include "gtest/gtest.h"
 
+using namespace mlir;
 using namespace mlir::tcp;
```
AaronStGeorge, line 16: Not important, but you could do `using namespace tcp`.

AaronStGeorge (approved):

PR #49, opened 2024-03-07
Run MathToLLVM before MathToLibm in tcp-to-llvm-pipeline

sjain-stanford:
Onboard `torch.sigmoid` to `aot_compile` and fix the tcp-to-llvm pipeline phase ordering (run `MathToLLVM` before `MathToLibm`). Without this any math ops remaining after `MathToLibm` fail legalization.


navahgar (approved):
LGTM

PR #50, opened 2024-03-08
Lower `aten.cat` -> `tensor::ConcatOp` and remove `tcp.concat`

sjain-stanford:
We try to leverage `tensor` dialect ops as much as possible if it fits our need. We don't seem to be needing a separate `tcp.concat`. This PR:
- Removes tcp.concat op definitions, verifiers, lit tests and TorchToTcp lowerings
- Update TorchToTcp conversion for `aten.concat` to use `tensor::ConcatOp` (adapted from TorchToLinalg)
- Adds pass to decompose tensor concat ops `DecomposeTensorConcatOpsPass`, which populates upstream [patterns](https://sourcegraph.com/github.com/llvm/llvm-project@main/-/blob/mlir/lib/Dialect/Tensor/Transforms/ConcatOpPatterns.cpp?L90:20-90:57) to decompose `tensor::ConcatOp` to `tensor::EmptyOp` and series of `tensor::InsertSliceOp`s. 
- Include `LowerAffinePass` and `ArithToLLVMConversionPass` in tcp -> llvm pipeline to handle the `affine.apply` ops resulting from the tensor.concat decomposition.
- Minor miscellaneous cleanup


lib/Conversion/TorchToTcp/DataMovement.cpp
```diff
@@ -28,21 +29,22 @@ using namespace mlir::torch;
...
 namespace {
-// Following function is copied from
-// https://sourcegraph.com/github.com/llvm/torch-mlir@4446fa00d8258311867496fc79d0b1dddd22a972/-/blob/lib/Conversion/TorchToLinalg/DataMovement.cpp?L42
-// TODO: Expose this function in a header to reuse.
-LogicalResult prepareArgumentsForSlicingOp(AtenSliceTensorOp op,
-                                           AtenSliceTensorOpAdaptor adaptor,
+// Adopted from
+// https://sourcegraph.com/github.com/llvm/torch-mlir@6b3a7d07c2c76f5e8437ff4e88110899621557b9/-/blob/lib/Conversion/TorchToLinalg/DataMovement.cpp?L42
+template <typename OpTy, typename OpAdaptor>
+LogicalResult prepareArgumentsForSlicingOp(OpTy op, OpAdaptor adaptor,
```
sjain-stanford, line 35: There were new updates to this function upstream - I've updated it here, and added the templatization back to make it easier to sync with upstream going forward.

include/mlir-tcp/Conversion/StablehloToTcp/StablehloToTcp.h
```diff
@@ -18,7 +19,7 @@ namespace mlir {
 
 namespace tcp {
 
-std::unique_ptr<Pass> createConvertStablehloToTcpPass();
+std::unique_ptr<OperationPass<func::FuncOp>> createConvertStablehloToTcpPass();
```
sjain-stanford, line 22: [Unrelated to this PR] 
Fix conversion pass creators to be explicit about returning function passes (`std::unique_ptr<OperationPass<func::FuncOp>>`)

lib/Dialect/Transforms/FuseTcpOpsPass.cpp
```diff
@@ -32,7 +33,8 @@ class TcpFuseElementwiseOpsPass
              use->hasTrait<OpTrait::Elementwise>();
     };
     patterns.add<GenericBottomUpFuser>(context, canFuse);
-    (void)applyPatternsAndFoldGreedily(op, std::move(patterns));
+    if (failed(applyPatternsAndFoldGreedily(op, std::move(patterns))))
+      return signalPassFailure();
```
sjain-stanford, line 37: (minor) Don't ignore failure in applying patterns.

tools/aot/aot_compile.bzl
```diff
@@ -242,7 +242,7 @@ def aot_compile(
                 "//tools/aot:abi",
                 "@cnpy//:cnpy",
                 "@com_google_googletest//:gtest_main",
-                "@llvm-project//mlir:mlir_c_runner_utils_hdrs",
+                "@llvm-project//mlir:mlir_c_runner_utils",
```
sjain-stanford, line 245: This is needed for `memrefCopy` symbol ([source](https://sourcegraph.com/github.com/llvm/llvm-project@762cbd82da4debf8b026a4eb4ade66720acf3182/-/blob/mlir/lib/ExecutionEngine/CRunnerUtils.cpp?L62:17-62:27)).

include/mlir-tcp/Dialect/Transforms/Passes.td
```diff
@@ -30,4 +30,11 @@ def VerifyTcpBackendContract : Pass<"torch-verify-tcp-backend-contract", "Module
   let constructor = "mlir::tcp::createVerifyTcpBackendContractPass()";
 }
 
+// \brief This pass decomposes `tensor::ConcatOp` into a `tensor::EmptyOp`
+// and a sequence of `tensor::InsertSliceOp`s.
+def DecomposeTensorConcatOps : Pass<"decompose-tensor-concat-ops", "func::FuncOp"> {
```
navahgar, line 35: I'm suspecting we will need to add more such transforms in future. So, can we name this pass something more generic, say, TransformTensorOps, so that we can accommodate more of these if necessary, in future.

navahgar (commented):
Thanks for taking this up. This looks much better.

Mostly LGTM. One high-level comment inlined below.

include/mlir-tcp/Dialect/Transforms/Passes.td
sjain-stanford: I've renamed the file to `TransformTensorOpsPass.h/cpp` to keep all tensor op related transforms colocated and minimizing the boilerplate, however chose to keep the pass name for decomposing `tensor.concat` the same for clarity, as I anticipate adding a new transform to the same file but register it as a separate / standalone pass. Does that sound reasonable? We can always reorganize once we uncover more such cases.

navahgar: Personally, I'd prefer having a single pass that covers all the transforms for Tensor dialect ops, because building a pipeline would require knowing all the passes, if we do them separately. 

But we can cross that bridge when we get there, as you said.

navahgar (approved):
LGTM

include/mlir-tcp/Dialect/Transforms/Passes.td
sjain-stanford: Renamed the pass to `DecomposeTensorOps` to make room of future tensor decompositions. For something other than a decomposition, I propose adding a separate pass (to the same cpp), and use a function in Pipeline.cpp that  creates a sequence of such "tensor transforms" passes.

PR #51, opened 2024-03-12
[AOT] Add integer concat to the suite

sjain-stanford:
In addition to testing integer concatenations, this also tests the integer type specializations in the aot_compile generated execute tests.


zezhang (approved):
LGTM! Thanks for working on the concat lowering.

PR #54, opened 2024-03-14
TcpToLinalg legality fixes: Mark only the Tcp ops with conversions as illegal

sjain-stanford:
TcpToLinalg conversion previously marked Tcp dialect as illegal, despite doing a partial conversion. So if there are any leftover Tcp ops after TcpToLinalg, they'd fail legalization (e.g. `tcp.const`). This is not correct and this partial conversion should only mark the Tcp ops it does convert as illegal. This PR fixes that.

### DIFF
--- a/lib/Conversion/StablehloToTcp/StablehloToTcp.cpp
+++ b/lib/Conversion/StablehloToTcp/StablehloToTcp.cpp
@@ -22,7 +22,7 @@
 
 namespace mlir {
 
-#define GEN_PASS_DEF_CONVERTTCPTOLINALG
+#define GEN_PASS_DEF_CONVERTSTABLEHLOTOTCP
 #include "mlir-tcp/Conversion/Passes.h.inc"
 
 namespace tcp {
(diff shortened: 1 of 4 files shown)


lib/Conversion/StablehloToTcp/StablehloToTcp.cpp
```diff
@@ -22,7 +22,7 @@
 
 namespace mlir {
 
-#define GEN_PASS_DEF_CONVERTTCPTOLINALG
+#define GEN_PASS_DEF_CONVERTSTABLEHLOTOTCP
```
navahgar, line 25: Wondering how it was working earlier.

navahgar (approved):
LGTM. Thanks for fixing this.

lib/Conversion/StablehloToTcp/StablehloToTcp.cpp
sjain-stanford: No idea. I'm thinking if it is even needed here.

sjain-stanford: If I were to guess, it has to do with using a unique define to avoid including tablegen generated headers multiple times... Not sure though...

PR #58, opened 2024-03-14
LIT test for `decompose-tensor-ops` pass

sjain-stanford:
The lit test was missed in https://github.com/cruise-automation/mlir-tcp/pull/50 which added the pass to decompose tensor concat ops and only included AOT e2e tests.


navahgar (approved):
LGTM

PR #61, opened 2024-03-20
[T-REX] Add more elementwise ops to AOT test-suite and bugfix for `tcp.sqrt` with integer inputs

sjain-stanford:
Adds the following ops to T-REX (Tcp Reference EXecution engine):

- scalar add, sub, mul, div
- sub tensor with alpha
- div tensor with mixed ranks
- tanh
- clamp
- relu (lowers to tcp.clamp)
- sqrt_int, sqrt_float

This exposed another bug in TcpToLinalg for lowering `tcp.sqrt` specifically for integer inputs. We were blanket converting `tcp.sqrt` to `math.sqrt` ([here](https://sourcegraph.com/github.com/cruise-automation/mlir-tcp@1da94c618220d16104627ac5af6145fc5b1a9022/-/blob/lib/Conversion/TcpToLinalg/Elementwise.cpp?L110)), but `math.sqrt` doesn't support integer inputs.
```
bazel-out/k8-fastbuild/bin/test/AotCompile/_internal_sqrt_int_tcp.mlir:3:10: error: 'math.sqrt' op operand #0 must be floating-point-like, but got 'i32'                                                                                                                                                                                                                         
    %0 = tcp.sqrt %arg0 : tensor<?x3xi32> -> tensor<?x3xf32>                                                                                                                                                                                                                                                                                                                     
         ^                                                                                                                                                                                                                                                                                                                                                                       
bazel-out/k8-fastbuild/bin/test/AotCompile/_internal_sqrt_int_tcp.mlir:3:10: note: see current operation: %4 = "math.sqrt"(%arg1) <{fastmath = #arith.fastmath<none>}> : (i32) -> i32                                                                                                                                                                                            
Target //test/AotCompile:sqrt_int_compile_execute_test failed to build
```

I considered two potential fixes:
1. Add `arith::SIToFPOp` cast when lowering tcp.sqrt in TcpToLinalg, or
1. Add `tcp::CastOp` in TorchToTcp and let our lowering to Linalg convert this to `arith::SIToFPOp` automatically.

I've implemented 2 because it brings tcp.sqrt consistent with the math dialect op (only support float inputs), and the various casting semantics is easier handled using a single tcp::Cast op rather than writing custom conversion to the plethora of arith cast ops, maximizing code reuse. This is also consistent with how TorchToArith handles it (see [test](https://sourcegraph.com/github.com/llvm/torch-mlir@0a581a97a759d34950d48aa7b03ab0a048b60259/-/blob/test/Conversion/TorchToArith/basic.mlir?L250:30-250:49)).


test/AotCompile/model_loader_lib.py
```diff
@@ -129,6 +129,92 @@ def forward(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
     )
 
 
+def sub_tensor_with_alpha_loader() -> TorchLoaderOutput:
+    class SubTensorWithAlpha(torch.nn.Module):
+        def __init__(self):
+            super().__init__()
+
+        def forward(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
+            add = torch.sub(x, y, alpha=2)
```
navahgar, line 138: nit: s/add/sub

test/AotCompile/model_loader_lib.py
```diff
@@ -129,6 +129,92 @@ def forward(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
...
+    )
+
+
+def div_tensor_mixed_ranks_loader() -> TorchLoaderOutput:
+    class DivTensorMixedRanks(torch.nn.Module):
+        def __init__(self):
+            super().__init__()
+
+        def forward(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
+            add = torch.div(x, y)
```
navahgar, line 165: nit: s/add/div

navahgar (approved):
LGTM. Thanks for adding tests for all these ops.

PR #62, opened 2024-03-21
[TorchToTcp] Don't broadcast axes that don't change size

AaronStGeorge:
Ran into this a bug in Cruse internal repo, the changes here fix the issue. Here's an example of incorrect lowering:
```mlir
  %4 = torch.aten.broadcast_to %3, %1 : !torch.vtensor<[1,2,1,2],f32>, !torch.list<int> -> !torch.vtensor<[4,2,4,2],f32>
```
was lowering to 
```mlir
%13 = tcp.broadcast %4, %6, %8, %10, %12 {axes = [0, 1, 2, 3]} : tensor<1x2x1x2xf32>, index, index, index, index -> tensor<4x2x4x2xf32>
```
axes 1 and 3 should not be broadcasted as those axes stay the same shape. The correct lowering is
```mlir
%13 = tcp.broadcast %4, %6, %10 {axes = [0, 2]} : tensor<1x2x1x2xf32>, index, index -> tensor<4x2x4x2xf32>
```

### DIFF
--- a/lib/Conversion/TorchToTcp/Misc.cpp
+++ b/lib/Conversion/TorchToTcp/Misc.cpp
@@ -95,18 +95,19 @@ public:
       int64_t staticDimSize;
       bool isDimDynamic =
           !matchPattern(newDimSize, m_TorchConstantInt(&staticDimSize));
-      // lambda because staticDimSize will contain garbage in the case of a
-      // dynamic dim.
-      auto isDimBroadcasted = [i, &inputShape,
-                               newLeadingDims](int64_t dimSize) -> bool {
-        // pytorch defines "passing -1 as the size for a dimension means not
-        // changing the size of that dimension."
-        bool isDimSizePreserved = dimSize == -1;
-        bool doesDimChangeShape = dimSize != inputShape[i - newLeadingDims];
+      // pytorch defines "passing -1 as the size for a dimension means not
+      // changing the size of that dimension." Short circuit if dim is dynamic
+      // as staticDimSize won't have a valid value.
+      bool isDimSizePreserved = isDimDynamic ? false : staticDimSize == -1;
+      // Short circuit if isNewDim to prevent out of bounds access of
+      // inputShape.
+      bool doesDimChangeShape =
+          isDimDynamic || isNewDim
+              ? false
+              : staticDimSize != inputShape[i - newLeadingDims];
 
-        return !isDimSizePreserved && doesDimChangeShape;
-      };
-      if (isNewDim || isDimDynamic || isDimBroadcasted(staticDimSize)) {
+      if (isNewDim || isDimDynamic ||
+          (!isDimSizePreserved && doesDimChangeShape)) {
         axes.push_back(i);
         newDimSize = rewriter.create<torch::TorchConversion::ToI64Op>(
             op->getLoc(), newDimSize);
--- a/lib/Conversion/TorchToTcp/Misc.cpp
+++ b/lib/Conversion/TorchToTcp/Misc.cpp
@@ -85,11 +85,28 @@ public:
     SmallVector<Value> resultShape;
     for (int64_t i = 0; i < static_cast<int64_t>(newDimSizes.size()); ++i) {
       Value newDimSize = newDimSizes[i];
+
+      bool isNewDim = i < newLeadingDims;
+      // At the pytorch level it's possible to broadcast to a dynamic dimension;
+      // for example: `torch.broadcast_to(x, y.size())` where `y.size()` has
+      // dynamic shapes. When broadcasting to dynamic size, matchPattern fails
+      // to read an int out of the MLIR Value. In the "usual" case
+      // (`torch.broadcast_to(x, (3, 3))`) matchPattern will succeed.
       int64_t staticDimSize;
-      if (i < newLeadingDims ||
-          !matchPattern(newDimSize, m_TorchConstantInt(&staticDimSize)) ||
-          (staticDimSize != -1 &&
-           staticDimSize != inputShape[i - newLeadingDims])) {
+      bool isDimDynamic =
+          !matchPattern(newDimSize, m_TorchConstantInt(&staticDimSize));
+      // lambda because staticDimSize will contain garbage in the case of a
+      // dynamic dim.
+      auto isDimBroadcasted = [i, &inputShape,
+                               newLeadingDims](int64_t dimSize) -> bool {
+        // pytorch defines "passing -1 as the size for a dimension means not
+        // changing the size of that dimension."
+        bool isDimSizePreserved = dimSize == -1;
+        bool doesDimChangeShape = dimSize != inputShape[i - newLeadingDims];
+
+        return !isDimSizePreserved && doesDimChangeShape;
+      };
+      if (isNewDim || isDimDynamic || isDimBroadcasted(staticDimSize)) {
         axes.push_back(i);
         newDimSize = rewriter.create<torch::TorchConversion::ToI64Op>(
             op->getLoc(), newDimSize);
--- a/lib/Conversion/TorchToTcp/Misc.cpp
+++ b/lib/Conversion/TorchToTcp/Misc.cpp
@@ -68,6 +68,8 @@ public:
     Value input = adaptor.getSelf();
     RankedTensorType inputType = input.getType().dyn_cast<RankedTensorType>();
 
+    ArrayRef<int64_t> inputShape = inputType.getShape();
+
     SmallVector<Value> newDimSizes;
     if (!getListConstructElements(op.getSize(), newDimSizes))
       return rewriter.notifyMatchFailure(
@@ -86,7 +88,8 @@ public:
       int64_t staticDimSize;
       if (i < newLeadingDims ||
           !matchPattern(newDimSize, m_TorchConstantInt(&staticDimSize)) ||
-          staticDimSize != -1) {
+          (staticDimSize != -1 &&
+           staticDimSize != inputShape[i - newLeadingDims])) {
         axes.push_back(i);
         newDimSize = rewriter.create<torch::TorchConversion::ToI64Op>(
             op->getLoc(), newDimSize);
--- a/lib/Conversion/TorchToTcp/Misc.cpp
+++ b/lib/Conversion/TorchToTcp/Misc.cpp
@@ -76,38 +76,44 @@ public:
           op, "Broadcasted shape must be a list of scalars");
 
     int64_t newLeadingDims = newDimSizes.size() - inputType.getRank();
-    if (newLeadingDims > 0) {
+    if (newLeadingDims > 0)
       input = torch_to_tcp::broadcastRankInLeadingDims(rewriter, input,
                                                        newLeadingDims);
-    }
 
     SmallVector<int64_t> axes;
     SmallVector<Value> resultShape;
     for (int64_t i = 0; i < static_cast<int64_t>(newDimSizes.size()); ++i) {
       Value newDimSize = newDimSizes[i];
 
+      // Per PyTorch, "Tensor can be also expanded to a larger number of
+      // dimensions, and the new ones will be appended at the front. For
+      // the new dimensions, the size cannot be set to -1.", so this dim is
+      // always broadcasted (no need to check for `isDimSizePreserved` below)
       bool isNewDim = i < newLeadingDims;
-      // At the pytorch level it's possible to broadcast to a dynamic dimension;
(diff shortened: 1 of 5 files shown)


lib/Conversion/TorchToTcp/Misc.cpp
```diff
@@ -84,7 +86,8 @@ class ConvertAtenBroadcastToOp : public OpConversionPattern<AtenBroadcastToOp> {
       int64_t staticDimSize;
       if (i < newLeadingDims ||
           !matchPattern(newDimSize, m_TorchConstantInt(&staticDimSize)) ||
-          staticDimSize != -1) {
+          (staticDimSize != -1 &&
+           staticDimSize != inputShape[i - newLeadingDims])) {
```
sjain-stanford, line 90: perhaps add a short comment above explaining the rationale

test/Conversion/TorchToTcp/misc.mlir
```diff
@@ -322,3 +322,22 @@ func.func @torch.aten.size.int(%arg0: !torch.vtensor<[?,?],f32>) -> () {
...
+// CHECK-LABEL:  @torch.aten.broadcast_to(
+// CHECK-SAME:   %[[ARG:.*]]: !torch.vtensor<[1,2,1,2],f32>) {
+// CHECK:        %[[TENSOR:.*]] = torch_c.to_builtin_tensor %[[ARG]] : !torch.vtensor<[1,2,1,2],f32> -> tensor<1x2x1x2xf32>
+// CHECK:        %[[CONSTANT:.*]] = torch.constant.int 4
+// CHECK:        %[[CAST0:.*]] = torch_c.to_i64 %[[CONSTANT]]
+// CHECK:        %[[BROADCAST_DIM0:.*]] = arith.index_cast %[[CAST0]] : i64 to index
+// CHECK:        %[[CAST1:.*]] = torch_c.to_i64 %[[CONSTANT]]
+// CHECK:        %[[BROADCAST_DIM1:.*]] = arith.index_cast %[[CAST1]] : i64 to index
+// CHECK:        %{{.*}} = tcp.broadcast %[[TENSOR]], %[[BROADCAST_DIM0]], %[[BROADCAST_DIM1]] {axes = [0, 2]} : tensor<1x2x1x2xf32>, index, index -> tensor<4x2x4x2xf32>
+func.func @torch.aten.broadcast_to(%arg0: !torch.vtensor<[1,2,1,2],f32>) -> () {
```
sjain-stanford, line 337: Thanks for adding the lit test (this was missed earlier). Can you please also onboard this to the `aot_compile` macro for e2e testing? I usually just have chat-GPT write me the pytorch model to produce this Torch dialect program, and feed that into the model loader.

It gave me this:
```python
class MyModel(torch.nn.Module):
    def __init__(self):
        super(MyModel, self).__init__()

    def forward(self, arg0):
        # Define the desired shape as a list
        shape = [4, 2, 4, 2]
        # Broadcast the input tensor to the desired shape
        output = arg0.expand(*shape)
        return output

# Example usage
model = MyModel()
sample_input = torch.randn(1, 2, 1, 2)  # Example input shape [1, 2, 1, 2]
output = model(sample_input)
print(output)
```

lib/Conversion/TorchToTcp/Misc.cpp
navahgar: +1 to this.

How about adding some lambda function that describes what it is actually checking, say something like `isDimBroadcasted` for the new condition you added?

lib/Conversion/TorchToTcp/Misc.cpp
```diff
@@ -84,7 +86,8 @@ class ConvertAtenBroadcastToOp : public OpConversionPattern<AtenBroadcastToOp> {
       int64_t staticDimSize;
       if (i < newLeadingDims ||
           !matchPattern(newDimSize, m_TorchConstantInt(&staticDimSize)) ||
-          staticDimSize != -1) {
+          (staticDimSize != -1 &&
```
sjain-stanford, line 89: nit: can we use `ShapedType::kDynamic` instead of `-1` here?
```
(staticDimSize != ShapedType::kDynamic &&
...
```

test/Conversion/TorchToTcp/misc.mlir
sjain-stanford: @AaronStGeorge I did it for you here: https://github.com/cruise-automation/mlir-tcp/pull/63 :)

AaronStGeorge: Thanks @sjain-stanford :) !

lib/Conversion/TorchToTcp/Misc.cpp
AaronStGeorge: 💯 didn't know about that

sjain-stanford: As discussed offline, please ignore since this condition is to check for `-1` in `torch.broadcast_to` or `torch.expand` which stands for "unchanged dim".

test/Dialect/misc.mlir
```diff
@@ -65,6 +65,22 @@ func.func @test_broadcast_axes_w_duplicates(%arg0 : tensor<?x1x?x1xf32>, %arg1 :
...
+  // expected-error@+1{{'tcp.broadcast' op failed to verify that attribute `axes` are in bounds}}
+  %0 = tcp.broadcast %arg0, %arg1, %arg2 {axes = [1, 100]} : tensor<?x1x?x1xf32>, index, index -> tensor<?x?x?x?xf32>
+  return %0 : tensor<?x?x?x?xf32>
+}
+
+// -----
+
+func.func @test_broadcast_axes_not_1(%arg0 : tensor<?x7x?x1xf32>, %arg1 : index, %arg2 : index) -> tensor<?x?x?x?xf32> {
+  // expected-error@+1{{'tcp.broadcast' op failed to verify that dimensions listed in attribute `axes` have a static size of `1`}}
+  %0 = tcp.broadcast %arg0, %arg1, %arg2 {axes = [1, 1]} : tensor<?x7x?x1xf32>, index, index -> tensor<?x?x?x?xf32>
```
sjain-stanford, line 78: Perhaps use `axes = [1, 3]` here as `[1, 1]` triggers another category of error (axes must not have any duplicates).

test/Conversion/TorchToTcp/misc.mlir
```diff
@@ -322,3 +322,46 @@ func.func @torch.aten.size.int(%arg0: !torch.vtensor<[?,?],f32>) -> () {
...
+  %int2 = torch.constant.int 2
+  %int4 = torch.constant.int 4
+  %1 = torch.prim.ListConstruct %int4, %int2, %int4, %int2 : (!torch.int, !torch.int, !torch.int, !torch.int) -> !torch.list<int>
+  %2 = torch.aten.broadcast_to %arg0, %1 : !torch.vtensor<[1,2,1,2],f32>, !torch.list<int> -> !torch.vtensor<[4,2,4,2],f32>
+  return
```
sjain-stanford, line 342: nit: `return %2 : !torch.vtensor<[4,2,4,2],f32>`

lib/Conversion/TorchToTcp/Misc.cpp
```diff
@@ -81,10 +83,28 @@ class ConvertAtenBroadcastToOp : public OpConversionPattern<AtenBroadcastToOp> {
...
+      bool isDimDynamic =
+          !matchPattern(newDimSize, m_TorchConstantInt(&staticDimSize));
+      // lambda because staticDimSize will contain garbage in the case of a
+      // dynamic dim.
+      auto isDimBroadcasted = [i, &inputShape,
+                               newLeadingDims](int64_t dimSize) -> bool {
+        // pytorch defines "passing -1 as the size for a dimension means not
+        // changing the size of that dimension."
+        bool isDimSizePreserved = dimSize == -1;
+        bool doesDimChangeShape = dimSize != inputShape[i - newLeadingDims];
```
navahgar, line 103: This will result in accessing with negative indices when `i < newLeadingDim`, if this lambda is used separately. It is probably not happening now because this lambda is preceded by `isNewDim` condition in your check below.

Since, this is now separated out, I suggest adding the `i >= newLeadingDims` check here as well. Wdyt?

sjain-stanford (commented):
Before landing, please rebase / merge main, and enable [this test](https://github.com/cruise-automation/mlir-tcp/blob/main/test/AotCompile/BUILD#L21) in CI.

lib/Conversion/TorchToTcp/Misc.cpp
sjain-stanford: How about we drop the lambda entirely, and do something like this instead, to handle both the negative indices issue and the garbage staticDimSize issue?
```cpp
bool isNewDim = i < newLeadingDims;
int64_t staticDimSize;
bool isDimDynamic = !matchPattern(newDimSize, m_TorchConstantInt(&staticDimSize));
bool isDimSizePreserved = isDimDynamic ? false : staticDimSize == -1;
bool doesDimSizeChange = isNewDim ? false : staticDimSize != inputShape[i - newLeadingDims];
if (isNewDim || isDimDynamic || (!isDimSizePreserved && doesDimSizeChange)) {
...
```

AaronStGeorge: `doesDimSizeChange` also requires a non dynamic dim. How about just a comment that says the order in which the booleans are checked matters here?

sjain-stanford: huh I realized that and edited it, but I think GH doesn't send edit emails 😅 
```cpp
bool doesDimSizeChange = (isNewDim || isDimDynamic) ? false : staticDimSize != inputShape[i - newLeadingDims];
```

navahgar (approved):
LGTM. Thanks for persevering and getting everything fixed properly here.

lib/Conversion/TorchToTcp/Misc.cpp
```diff
@@ -83,10 +85,29 @@ class ConvertAtenBroadcastLikeOps : public OpConversionPattern<AtenOpT> {
...
+      // At the pytorch level it's possible to broadcast to a dynamic dimension;
+      // for example: `torch.broadcast_to(x, y.size())` where `y.size()` has
+      // dynamic shapes. When broadcasting to dynamic size, matchPattern fails
+      // to read an int out of the MLIR Value. In the "usual" case
+      // (`torch.broadcast_to(x, (3, 3))`) matchPattern will succeed.
       int64_t staticDimSize;
-      if (i < newLeadingDims ||
-          !matchPattern(newDimSize, m_TorchConstantInt(&staticDimSize)) ||
-          staticDimSize != -1) {
+      bool isDimDynamic =
```
sjain-stanford, line 96: nit: prefer `isDynamicDim` for consistency with `isNewDim`.

lib/Conversion/TorchToTcp/Misc.cpp
```diff
@@ -83,10 +85,29 @@ class ConvertAtenBroadcastLikeOps : public OpConversionPattern<AtenOpT> {
     SmallVector<Value> resultShape;
     for (int64_t i = 0; i < static_cast<int64_t>(newDimSizes.size()); ++i) {
       Value newDimSize = newDimSizes[i];
+
+      bool isNewDim = i < newLeadingDims;
```
sjain-stanford, line 89: Please add a comment here clarifying PyTorch semantics:
> "Tensor can be also expanded to a larger number of dimensions, and the new ones will be appended at the front. For the new dimensions, the size cannot be set to -1."

lib/Dialect/IR/TcpOps.cpp
```diff
@@ -49,6 +49,22 @@ LogicalResult BroadcastOp::verify() {
...
+    if (axis >= inputShape.size()) {
+      return emitOpError(
+          "failed to verify that attribute `axes` are in bounds");
+    }
+
+    if (inputShape[axis] != 1) {
+      return emitOpError("failed to verify that dimensions listed in attribute "
+                         "`axes` have a static size of `1`");
+    }
+  }
```
sjain-stanford, line 66: Thanks for adding these checks to the verifier!

sjain-stanford (approved):
Nice catch and thanks for addressing verifier gaps. LGTM besides a few minor naming/documentation requests. 

[non-blocking] In a follow-on , I think we should include tests for the case of `torch.expand` or `torch.broadcast_to` adding new dims (increasing rank).

sjain-stanford:
@AaronStGeorge just saw your vacation ping - please do not bother, I can take over and land this after addressing any remaining nits.

PR #68, opened 2024-04-25
Add op for Gather

navahgar:
This PR adds the `tcp.gather` op. The design and semantics of this op are defined in [docs/gather.md](https://github.com/cruise-automation/mlir-tcp/blob/main/docs/gather.md).

More specifically, this PR includes:
* The definition of the `tcp.gather` op.
* Lowering from Torch to Tcp for the `torch.gather` op.
* Lowering from Tcp to Linalg for the `tcp.gather` op.
* Lit tests for the op and the lowerings.
* AOT test the Tcp gather op.
* Renames the op in the design document.

This PR does not update the lowering we have from aten.gather to tcp.custom("aten.gather") pattern [here](https://github.com/cruise-automation/mlir-tcp/blob/b812717001121df6974687e2b2e7e82b900390dd/lib/Conversion/TorchToTcp/TcpCustomOp.cpp#L29-L45). We should be able to control how `aten.gather` gets lowered by fixing the order in which the conversion passes are called in the pipeline.

### DIFF
--- a/lib/Conversion/TcpToLinalg/DataMovement.cpp
+++ b/lib/Conversion/TcpToLinalg/DataMovement.cpp
@@ -1,0 +1,101 @@
+//===------------------------------------------------------------*- C++ -*-===//
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// Also available under a BSD-style license. See LICENSE.
+//
+//===----------------------------------------------------------------------===//
+
+#include "mlir-tcp/Conversion/TcpToLinalg/TcpToLinalg.h"
+
+#include "mlir-tcp/Dialect/IR/TcpDialect.h"
+#include "mlir-tcp/Dialect/IR/TcpOps.h"
+
+#include "../PassDetail.h"
+#include "PopulatePatterns.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Math/IR/Math.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Transforms/DialectConversion.h"
+
+#include "llvm/Support/Debug.h"
+
+using namespace mlir;
+using namespace mlir::tcp;
+
+namespace {
+
+class ConvertGatherOp : public OpConversionPattern<GatherOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(GatherOp op, OpAdaptor adaptor,
+                                ConversionPatternRewriter &b) const override {
+    Location loc = op->getLoc();
+    auto resultTensorType = getTypeConverter()
+                                ->convertType(op->getResult(0).getType())
+                                .cast<RankedTensorType>();
+
+    auto inputTensor = op->getOperands()[0];
+    auto indicesTensor = op->getOperands()[1];
+
+    int64_t gatherDim = op.getDimAttr().getValue().getSExtValue();
+
+    auto resultRank = resultTensorType.getRank();
+
+    SmallVector<Value> resultDimSizes;
+    for (int64_t i = 0; i < resultRank; ++i) {
+      resultDimSizes.push_back(
+          b.createOrFold<tensor::DimOp>(loc, indicesTensor, i));
+    }
+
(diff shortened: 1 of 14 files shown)


test/AotCompile/model_loader_lib.py
navahgar, line 1: All the changes in this file except the `gather_elements_loader` are formatting changes.

lib/Conversion/TcpToLinalg/DataMovement.cpp
```diff
@@ -0,0 +1,101 @@
...
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(GatherOp op, OpAdaptor adaptor,
+                                ConversionPatternRewriter &b) const override {
+    Location loc = op->getLoc();
+    auto resultTensorType = getTypeConverter()
+                                ->convertType(op->getResult(0).getType())
+                                .cast<RankedTensorType>();
+
+    auto inputTensor = op->getOperands()[0];
```
srinathava, line 42: nit: Use the strongly typed methods like `op.getInput` and `op.getIndices` instead of `op->getOperands()[0]`

srinathava, line 42: I've also never quite understood whether its safe to use `op->...` methods or one should use `adaptor.getInput` to account for prior replacements. Might be cleaner to always use `adaptor.getInput` etc. here?

lib/Conversion/TcpToLinalg/DataMovement.cpp
```diff
@@ -0,0 +1,101 @@
...
+    indexingMaps.push_back(b.getMultiDimIdentityMap(resultRank));
+
+    SmallVector<utils::IteratorType> iteratorTypes(
+        resultRank, utils::IteratorType::parallel);
+
+    Value emptyTensor =
+        b.create<tensor::EmptyOp>(loc, getAsOpFoldResult(resultDimSizes),
+                                  resultTensorType.getElementType());
+
+    auto bodyBuilder = [&](OpBuilder &b, Location loc, ValueRange payloadArgs) {
```
srinathava, line 66: nit: slightly confusing to have the same named variable `b` for both the outer rewriter and the OpBuilder here..

srinathava (commented):
Looks good to me... Mostly just nits apart from the question about whether we should be using the OpAdaptor in the matchAndRewrite.

lib/Conversion/TcpToLinalg/DataMovement.cpp
zezhang: My understanding is that` adaptor.getInput` gets the input after all conversions, whereas `op.getInput` gets the original input.

e.g. assuming we have

1% = tcp.sigmoid(xxx)
2% =  tcp.exp(%1)

which lowers to 

1% = linalg.sigmoid(xxx)
2% =  linalg.exp(%1)

For the `exp` op, if we do op.getInput, we will get the value of `tcp.sigmoid`, but adaptor.getInput will get the value of `linalg.sigmod`, which is after other conversions if there is any.

zezhang (commented):
Should we also remove the aten.gather --> tcp.custom("aten.gather") pattern [here](https://github.com/cruise-automation/mlir-tcp/blob/b812717001121df6974687e2b2e7e82b900390dd/lib/Conversion/TorchToTcp/TcpCustomOp.cpp#L29-L45)?

Cruise internal repos have some deps on `tcp.custom("aten.gather")` and further lowers it to a run-time kernel. May also need to verify these changes from internal repos.

lib/Conversion/TcpToLinalg/DataMovement.cpp
navahgar: Good point. Updated the outer variable to `rewriter`.

lib/Conversion/TcpToLinalg/DataMovement.cpp
navahgar: Sure. Done.

navahgar: That is my understanding too. This makes a big difference especially when the types change during the conversion. In this case, it shouldn't make much of a difference between the two usages.

However, I agree that it is better to use `adaptor` based methods for clarify and consistency. Updated this code.

navahgar (commented):
Thanks for the reviews @srinathava and @zezhang. Addressed your comments PTAL.

navahgar:
> Cruise internal repos have some deps on tcp.custom("aten.gather") and further lowers it to a run-time kernel. May also need to verify these changes from internal repos.

This is one of the reasons I don't want to remove this custom op lowering. We can control how this op gets lowered by ordering the passes in the pipeline. We can clean this up later if needed. 

Updated the PR description to mention this.

srinathava (approved):
LGTM!

PR #70, opened 2024-05-14
[TorchToTcp] Mark torch ops with non-fp inputs dynamically legal

sjain-stanford:
This helps avoid legalization errors after Torch -> Tcp, for ops like `aten.neg` and `aten.sigmoid` with integer inputs, that do not convert to Tcp (due to type restrictions) and fail legalization, preventing further dialect conversions (to Tosa for instance).


zezhang (approved):
Thanks for working on this. LGTM!

PR #71, opened 2024-05-16
Remove index array type

mabubakarpurdue:
Remove index array type since it is not being used


sjain-stanford (approved):

sjain-stanford:
Please also remove the following unused code from `TcpBase.td`:
```
// Tcp Types Base

class Tcp_Type<string name, string typeMnemonic> : TypeDef<Tcp_Dialect, name> {
    let mnemonic = typeMnemonic;
}
```

PR #72, opened 2024-06-12
Add tcp.slice op

navahgar:
This PR adds the `tcp.slice` op and the conversions needed for that op.

Specifically, this PR:
* Adds a new op in the Tcp dialect, `tcp.slice`.
* Updates the lowering of slice in TorchToTcp to create `tcp.slice`. Prior to this change, this lowering was emitting `tensor.extract_slice` instead.
* Adds a new lowering pass, `TcpToTensor`, which lowers `tcp.slice` to `tensor.extract_slice`.
* Adds unit tests and conversion tests for both the lowerings.
* Adds an e2e execution test that uses `tcp.slice`.

### DIFF
--- a/include/mlir-tcp/Dialect/IR/TcpOps.td
+++ b/include/mlir-tcp/Dialect/IR/TcpOps.td
@@ -618,4 +618,26 @@ def Tcp_GatherOp : Tcp_Op<"gather", [Pure, AllElementTypesMatch<["input", "out"]
   let assemblyFormat = "$input `,` $indices attr-dict `:` type($input) `,` type($indices) `->` type($out)";
 }
 
+def Tcp_SliceOp : Tcp_Op<"slice", [Pure, AllElementTypesMatch<["in", "out"]>, SameVariadicOperandSize]> {
+
+  let summary = "Extracts a slice of the input tensor";
+
+  let description = [{
+    Extracts a slice of the input tensor based on the given starts, sizes, and strides.
+  }];
+
+  let arguments = (ins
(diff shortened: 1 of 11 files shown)


include/mlir-tcp/Dialect/IR/TcpOps.td
```diff
@@ -618,4 +618,26 @@ def Tcp_GatherOp : Tcp_Op<"gather", [Pure, AllElementTypesMatch<["input", "out"]
...
+    Variadic<Index>:$starts,
+    Variadic<Index>:$sizes,
+    Variadic<Index>:$strides
+  );
+
+  let results = (outs
+    Tcp_Tensor:$out
+  );
+
+  let assemblyFormat = "$in `(` $starts `)` `(` $sizes `)` `(` $strides `)` attr-dict `:` type($in) `->` type($out)";
```
srinathava, line 640: minor nit: would it make sense for the assembly format to hint at what each group is for? So something like:

```
let assemblyFormat = "$in `starts(` $starts `)` `, sizes(` $sizes `)` `, strides(` $strides `)` attr-dict `:` type($in) `->` type($out)";
```

srinathava (approved):
LGTM apart from a minor nit.

include/mlir-tcp/Dialect/IR/TcpOps.td
navahgar: Thats a good idea. It definitely makes the op more readable. Thanks for the suggestion. Added

PR #74, opened 2024-06-21
Allow creation of TCP groups where an op has multiple uses

srinathava:
We previously only allowed an op to have a single use during the creation a group for it. This PR relaxes that to allow multiple uses as long as all the uses belong to the same region.


lib/Dialect/Transforms/FusionPatterns.cpp
```diff
@@ -12,70 +12,145 @@
...
+  // Currently we are only fusing ops at the top-level.
+  // This is to avoid recursing inside a group and ending up with
+  // nested groups that contain the same ops.
+  // Since we are iterating bottom up in a block, we only need to check
+  // if the def op has a func parent.
+  //
+  // TODO: Remove this restriction to allow fusing in nested regions.
+  if (!isa<func::FuncOp>(op->getParentOp())) {
+    return failure();
+  }
```
sjain-stanford, line 38: style nit: prefer omitting braces for single-statement ifs ([reference](https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements))

lib/Dialect/Transforms/FusionPatterns.cpp
```diff
@@ -12,70 +12,145 @@
...
+  // same region and can be fused with the defining op
+  Region *usesParentRegion = nullptr;
+  SmallVector<Operation *> uses;
+  llvm::DenseSet<Operation *> usesSet;
+
+  for (auto &use : op->getUses()) {
+    auto parentRegion = use.getOwner()->getParentRegion();
+    if (usesParentRegion && usesParentRegion != parentRegion) {
+      return failure();
+    }
```
sjain-stanford, line 53: same

lib/Dialect/Transforms/FusionPatterns.cpp
```diff
@@ -12,70 +12,145 @@
...
+    usesParentRegion = parentRegion;
+
+    if (!canFuse(op, use.getOwner())) {
+      return failure();
+    }
+    if (usesSet.insert(use.getOwner()).second) {
+      uses.push_back(use.getOwner());
+    }
```
sjain-stanford, line 61: same

lib/Dialect/Transforms/FusionPatterns.cpp
```diff
@@ -12,70 +12,145 @@
...
+      }
+      op->moveBefore(*uses.begin());
+    }
+  } else if (auto groupOp =
+                 dyn_cast<tcp::GroupOp>(usesParentRegion->getParentOp())) {
+    // Given that we iterate over the funcop in a bottom up manner, when moving
+    // into an existing group, we would be guaranteed that this op does not use
+    // any of the ops already in the group. So we can move it to the very
+    // beginning of the group. This ensures that the order of operands is
+    // preserved when creating a gruop. Otherwise, if we start with
```
sjain-stanford, line 119: typo: `group`

sjain-stanford, line 119: nit: IIUC you probably want `s/Otherwise/For example` here?

test/Dialect/tcp_fusion.mlir
```diff
@@ -51,3 +51,43 @@ func.func @test_multiple_fusions(%arg0 : tensor<?x?xf32>,
...
+
+// CHECK:   func.func @test_multi_use_fusion(%[[ARG0:.+]]: tensor<?x?xf32>, %[[ARG1:.+]]: tensor<?x?xf32>) -> (tensor<?x?xf32>, tensor<?x?xf32>) {
+// CHECK:     %[[V0:.+]]:2 = tcp.group {
+// CHECK:       %[[V1:.+]] = tcp.tanh %[[ARG0]] : tensor<?x?xf32> -> tensor<?x?xf32>
+// CHECK:       %[[V2:.+]] = tcp.add %[[V1]], %[[V1]] : tensor<?x?xf32>, tensor<?x?xf32> -> tensor<?x?xf32>
+// CHECK:       %[[V3:.+]] = tcp.mul %[[V2]], %[[ARG1]] : tensor<?x?xf32>, tensor<?x?xf32> -> tensor<?x?xf32>
+// CHECK:       %[[V4:.+]] = tcp.sub %[[V2]], %[[ARG1]] : tensor<?x?xf32>, tensor<?x?xf32> -> tensor<?x?xf32>
+// CHECK:       tcp.yield %[[V3]], %[[V4]] : tensor<?x?xf32>, tensor<?x?xf32>
+// CHECK:     } : tensor<?x?xf32>, tensor<?x?xf32>
+// CHECK:     return %[[V0]]#1, %[[V0]]#0 : tensor<?x?xf32>, tensor<?x?xf32>
```
sjain-stanford, line 85: Any idea why the order is swapped here? 
Should it be `return %[[V0]]#0, %[[V0]]#1 : tensor<?x?xf32>, tensor<?x?xf32>` instead?

test/Dialect/tcp_fusion.mlir
```diff
@@ -51,3 +51,43 @@ func.func @test_multiple_fusions(%arg0 : tensor<?x?xf32>,
...
+// CHECK:     %[[V0:.+]] = tcp.group {
+// CHECK:       %[[V1:.+]] = tcp.tanh %[[ARG0]] : tensor<?x?xf32> -> tensor<?x?xf32>
+// CHECK:       %[[V2:.+]] = tcp.add %[[V1]], %[[ARG1]] : tensor<?x?xf32>, tensor<?x?xf32> -> tensor<?x?xf32>
+// CHECK:       %[[V3:.+]] = tcp.sub %[[V2]], %[[ARG1]] : tensor<?x?xf32>, tensor<?x?xf32> -> tensor<?x?xf32>
+// CHECK:       %[[V4:.+]] = tcp.mul %[[V2]], %[[V3]] : tensor<?x?xf32>, tensor<?x?xf32> -> tensor<?x?xf32>
+// CHECK:       tcp.yield %[[V4]] : tensor<?x?xf32>
+// CHECK:     } : tensor<?x?xf32>
+// CHECK:     return %[[V0]] : tensor<?x?xf32>
+// CHECK:   }
+func.func @test_multi_use_fusion(%arg0 : tensor<?x?xf32>, %arg1 : tensor<?x?xf32>) -> tensor<?x?xf32> {
```
sjain-stanford, line 67: nit: call this something different from below (is this single use fusion?)

sjain-stanford (approved):
Thanks for relaxing this constraint. LGTM, besides minor style nits and a clarification question.

test/Dialect/tcp_fusion.mlir
srinathava: I added some comments explaining what the test intent is. These are both for testing fusion with multiple uses. The difference is in whether the multiple uses are already part of a tcp group or whether we are creating a new group from the multiple uses for the first time.

sjain-stanford (approved):

PR #75, opened 2024-06-24
[Clangd] Add TcpToTensor to clangd targets

sjain-stanford:
This was recently added (by https://github.com/cruise-automation/mlir-tcp/pull/72). In the future, we'll look into automating the target list generation (possibly a GitHub Action).


navahgar (approved):
Thanks for catching this.

PR #77, opened 2024-06-25
Tiling interface for slice op

navahgar:
This PR adds an implementation of `TilingInterface` for the `tcp.slice` op. This is necessary to enable tile and fuse with `tcp.slice`.

### DIFF
--- a/include/mlir-tcp/Dialect/IR/TcpTilingInterfaceImpl.h
+++ b/include/mlir-tcp/Dialect/IR/TcpTilingInterfaceImpl.h
@@ -1,0 +1,12 @@
+#pragma once
+
+namespace mlir {
+
+class DialectRegistry;
+
+namespace tcp {
+
+void registerTilingInterfaceExternalModels(DialectRegistry &registry);
+
+} // namespace tcp
+} // namespace mlir
--- a/include/mlir-tcp/Dialect/IR/TcpTilingInterfaceImpl.h
+++ b/include/mlir-tcp/Dialect/IR/TcpTilingInterfaceImpl.h
@@ -1,3 +1,12 @@
+//===------------------------------------------------------------*- C++ -*-===//
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// Also available under a BSD-style license. See LICENSE.
+//
+//===----------------------------------------------------------------------===//
+
 #pragma once
 
 namespace mlir {
--- a/lib/Dialect/IR/TcpTilingInterfaceImpl.cpp
+++ b/lib/Dialect/IR/TcpTilingInterfaceImpl.cpp
@@ -19,21 +19,44 @@ using namespace mlir::tcp;
 
 namespace {
 
+Value getOpFoldResultAsValue(OpFoldResult ofr, OpBuilder &b,
+                             mlir::Location loc) {
+  // Case 1: Check for Value.
+  if (auto val = llvm::dyn_cast_if_present<Value>(ofr)) {
+    return val;
+  }
+  // Case 2: Check for IntegerAttr.
+  Attribute attr = llvm::dyn_cast_if_present<Attribute>(ofr);
+  auto intAttr = dyn_cast_or_null<IntegerAttr>(attr);
+  assert(intAttr && "Expected to find an integer in OpFoldResult");
+  return b.create<arith::ConstantIndexOp>(loc,
+                                          intAttr.getValue().getSExtValue());
+}
+
+SmallVector<Value> getOpFoldResultsAsValues(ArrayRef<OpFoldResult> ofrs,
+                                            OpBuilder &b, mlir::Location loc) {
+  SmallVector<Value> vals = llvm::map_to_vector(ofrs, [&](OpFoldResult ofr) {
+    return getOpFoldResultAsValue(ofr, b, loc);
+  });
+  return vals;
+}
+
 struct SliceOpTiling
     : public TilingInterface::ExternalModel<SliceOpTiling, tcp::SliceOp> {
 
   SmallVector<utils::IteratorType> getLoopIteratorTypes(Operation *op) const {
-    // The iterator types describe the type of the loops in the
-    // loop-nest for the eventual tiled op. For the slice op, these are
-    // just parallel loops.
     auto sliceOp = cast<tcp::SliceOp>(op);
     return SmallVector(sliceOp.getType().getRank(),
                        utils::IteratorType::parallel);
   }
 
   SmallVector<Range> getIterationDomain(Operation *op, OpBuilder &b) const {
-    // The iteration domain describes the loop bounds for the tiled
-    // operation.
+    // The iteration domain describes the loop bounds for the tiled operation.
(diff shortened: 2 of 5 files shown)


include/mlir-tcp/Dialect/IR/TcpTilingInterfaceImpl.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
```
sjain-stanford, line 1: fly by comment: Please include the license header here.

lib/Dialect/IR/TcpTilingInterfaceImpl.cpp
```diff
@@ -0,0 +1,164 @@
+#include "mlir-tcp/Dialect/IR/TcpTilingInterfaceImpl.h"
```
sjain-stanford, line 1: same

lib/Dialect/IR/TcpTilingInterfaceImpl.cpp
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    auto extractOp = b.create<tensor::ExtractSliceOp>(
+        loc, sliceOp.getIn(), newOffsets, newSizes, newStrides);
+
+    // Add a `tcp.slice` op on the tile.
+    auto zero = b.create<arith::ConstantIndexOp>(loc, 0);
+    SmallVector<Value> zeroOffsets(offsets.size(), zero.getResult());
+    auto sizeIntValues = getConstantIntValues(sizes);
+    assert(
```
srinathava, line 132: I don't know if this assert would work... What happens if we run `structured.fuse` on back to back `tcp.slice` ops with non-trivial strides for instance? Given that the `tensor.extract_slice` created above has a computed value for the size, this is the size which will be passed to the upstream `tcp.slice`'s tiling interface. Maybe you can give such an example a shot?

If we indeed do not want to support non-constant values for the sizes, this should probably be a condition for which we return failure() at the beginning of the function.

Also, I think all we want to do here is to create a tensor type where the corresponding dim is `kDynamic` for the case when the corresponding size is not a constant integer.

test/Dialect/structured_fuse.mlir
```diff
@@ -0,0 +1,127 @@
...
+
+    return %ret : tensor<32x16xf32>
+}
+
+module attributes {transform.with_named_sequence} {
+  transform.named_sequence @__transform_main(%arg0: !transform.any_op {transform.readonly}) {
+    %slice = transform.structured.match ops{["tcp.slice"]} in %arg0 : (!transform.any_op) -> !transform.any_op
+    %unary = transform.structured.match ops{["linalg.elemwise_unary"]} in %arg0 : (!transform.any_op) -> !transform.any_op
+
+    %1, %loops:2 = transform.structured.fuse %unary {tile_sizes = [1, 1], tile_interchange = [0, 1]}
```
srinathava, line 116: I think it might be good to either modify this test or add one more test where the `tile_sizes` are not `[1, 1]`. That will ensure that we stress the case of the non-unit stride a bit more. Might also be good to have a case where we have back to back `tcp.slice` ops with non-trivial strides and a non-trivial tile size just to be additionally safe.

lib/Dialect/IR/TcpTilingInterfaceImpl.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    // Offset on the input of slice is computed by:
+    //     start + offset_on_output * stride
+    //
+    // Tile and fuse algorithm does not work with stride != 1.
+    // In order to get around this, we extract a contiguous chunk with
+    // `tensor.extract_slice` and use the following `tcp.slice` op to
+    // extract the strided parts. So, the size on the
+    // `tensor.extract_slice` will be
+    //     size * stride
```
srinathava, line 112: Actually, now that I think of it, the contiguous size can be slightly smaller right? For instance, if the final slice `size` is just 1 element for a `tcp.slice` with a stride of 10, we'd still want to only extract 1 element given that we are already accounting for the correct starting offset. So the contiguous size can be:

```
newSize = (size - 1)*stride + 1
```

include/mlir-tcp/Dialect/IR/TcpTilingInterfaceImpl.h
navahgar: Done

lib/Dialect/IR/TcpTilingInterfaceImpl.cpp
navahgar: done

lib/Dialect/IR/TcpTilingInterfaceImpl.cpp
navahgar: Good point. That is not necessary. Changed it to use `kDynamic` when these are not constants.

test/Dialect/structured_fuse.mlir
navahgar: Added tests for these 2 cases

lib/Dialect/IR/TcpTilingInterfaceImpl.cpp
navahgar: I agree that will be slightly smaller. But would doing 2 extra operations (a sub and a add) be better than extracting the remaining elements as well? I'm not sure if that would have any benefits.

We can probably try this if needed later, if we see value in it. Wdyt?

lib/Dialect/IR/TcpTilingInterfaceImpl.cpp
```diff
@@ -0,0 +1,181 @@
...
+#include "llvm/Support/Debug.h"
+
+#include <numeric>
+
+using namespace mlir;
+using namespace mlir::tcp;
+
+namespace {
+
+Value getOpFoldResultAsValue(OpFoldResult ofr, OpBuilder &b,
```
srinathava, line 32: This and its sibling below already exist upstream. See here:

https://sourcegraph.robot.car/github.robot.car/cruise/mla-robocomp-llvm-project/-/blob/mlir/include/mlir/Dialect/Arith/Utils/Utils.h?L36

lib/Dialect/IR/TcpTilingInterfaceImpl.cpp
srinathava: Sounds good. Maybe a little comment saying something to that effect?

test/Dialect/structured_fuse.mlir
srinathava: Thank you!

srinathava (approved):
LGTM. Just a few nits.

lib/Dialect/IR/TcpTilingInterfaceImpl.cpp
navahgar: Added a comment

lib/Dialect/IR/TcpTilingInterfaceImpl.cpp
navahgar: Thanks for pointing that out. I found some utils related to `OpFoldResult` in https://sourcegraph.com/github.com/llvm/llvm-project@main/-/blob/mlir/include/mlir/Dialect/Utils/StaticValueUtils.h, but didn't realize there are more of these under `arith`. Wonder why they are separate.

PR #78, opened 2024-07-03
Symbolic shape expressions in TCP dialect

sjain-stanford:
Support for TorchDynamo captured symbolic shape expressions was added to Torch-MLIR recently (https://github.com/llvm/torch-mlir/pull/3372). This PR continues the work to lower these to TCP.

- [x] TCP op definitions, custom printer/parser/verifier
- [x] Dialect and conversion lit tests
- [x] Custom op python lit tests
- [x] Cleanup pass to remove bind ops

### DIFF
--- a/include/mlir-tcp/Dialect/IR/TcpOps.td
+++ b/include/mlir-tcp/Dialect/IR/TcpOps.td
@@ -11,6 +11,7 @@
 #define TCP_OPS
 
 include "mlir/IR/OpBase.td"
+include "mlir/IR/BuiltinAttributes.td"
 
 include "mlir-tcp/Dialect/IR/TcpBase.td"
 include "mlir-tcp/Dialect/IR/TcpEnums.td"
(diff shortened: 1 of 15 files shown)


include/mlir-tcp/Dialect/IR/TcpOps.td
```diff
@@ -640,4 +641,69 @@ def Tcp_SliceOp : Tcp_Op<"slice", [Pure, AllElementTypesMatch<["in", "out"]>, Sa
...
+    Example:
+    ```
+    tcp.bind_symbolic_shape %arg0, [%0, %1], affine_map<()[s0, s1] -> (s0, s1, 3)> : tensor<?x?x3xf32>
+    tcp.bind_symbolic_shape %out0, [%0, %1, %2], affine_map<()[s0, s1, s2] -> (s0, s1 * 2 + s2, 3)> : tensor<?x?x3xf32>
+    ```
+  }];
+  let arguments = (ins
+    Tcp_Tensor:$operand,
+    Variadic<AnySignlessInteger>:$shape_symbols,
+    Builtin_AffineMapAttr:$shape_expressions
```
matthewfl, line 702: we were using semi-affine iirc right?  is this the right type?

matthewfl: nvm, this looks like the same type you used in torch-mlir https://github.com/llvm/torch-mlir/pull/3372/files#diff-04b7c11362e098c211459acf50282f7364c65e6755720898182992b5a6a91c94R1397

lib/Dialect/Transforms/DropSymbolicShapeOpsPass.cpp
```diff
@@ -0,0 +1,61 @@
...
+namespace {
+
+class RemoveBindSymbolicShapeOps
+    : public OpRewritePattern<tcp::BindSymbolicShapeOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(tcp::BindSymbolicShapeOp op,
+                                PatternRewriter &rewriter) const override {
+    if (!op->getUsers().empty())
```
matthewfl, line 35: Is this ever true?
Isn't users going to be always empty as this op is used like `tcp.bind_symbolic_shape %arg0, [%0], affine_map<()[s0] -> (s0)>` where the returned value is ignored

include/mlir-tcp/Dialect/IR/TcpOps.td
jain-sambhaav: Hmm I didn't find a separate built-in type for `SemiAffineMapAttr`, and was under the impression`Builtin_AffineMapAttr` has the extensions for semi-affine expressions as well. In fact I did test this with `mod` expressions ([here](https://github.com/llvm/torch-mlir/blob/main/test/python/fx_importer/sympy_to_affine_expr_test.py#L62-L63)) which worked out fine.

sjain-stanford: Hmm I didn't find a separate built-in type for `SemiAffineMapAttr`, and was under the impression`Builtin_AffineMapAttr` has the extensions for semi-affine expressions as well. In fact I have a test with `mod` expressions ([here](https://github.com/llvm/torch-mlir/blob/main/test/python/fx_importer/sympy_to_affine_expr_test.py#L62-L63)) which worked fine.

lib/Dialect/Transforms/DropSymbolicShapeOpsPass.cpp
sjain-stanford: Yes this should never be true; I wanted an assertion here instead of a match failure, but I'll just remove this since the default dialect verifier should suffice.

lib/Dialect/Transforms/DropSymbolicShapeOpsPass.cpp
```diff
@@ -0,0 +1,61 @@
...
+};
+
+class DropSymbolicShapeOpsPass
+    : public DropSymbolicShapeOpsBase<DropSymbolicShapeOpsPass> {
+  void runOnOperation() override {
+    Operation *op = getOperation();
+    MLIRContext *context = op->getContext();
+    RewritePatternSet patterns(context);
+
+    patterns.add<RemoveBindSymbolicShapeOps>(context);
```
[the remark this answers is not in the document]
sjain-stanford: Correct, the `Pure` trait lets them get canonicalized away when there are no uses. So once this pass removes the `tcp.bind_symbolic_shape` ops, the `tcp.symbolic_int` ops go away automatically.

PR #80, opened 2024-07-09
Lower torch.aten.fake_quantize ops to tcp.custom_op

zezhang:
To test:

`bazel test //...`

### DIFF
--- a/lib/Conversion/TorchToTcp/TcpCustomOp.cpp
+++ b/lib/Conversion/TorchToTcp/TcpCustomOp.cpp
@@ -154,10 +154,10 @@ public:
     helper.addIntAttr("quant_max", op.getQuantMax());
 
     // scale
-    if (!op.getScale().getDefiningOp())
+    auto scaleOp = op.getScale().getDefiningOp();
+    if (!scaleOp)
       return rewriter.notifyMatchFailure(op, "Missing scale operation");
-    auto scaleTensor = dyn_cast<torch::Torch::ValueTensorLiteralOp>(
-        op.getScale().getDefiningOp());
+    auto scaleTensor = dyn_cast<torch::Torch::ValueTensorLiteralOp>(scaleOp);
     if (!scaleTensor)
       return rewriter.notifyMatchFailure(
           op, "Scale operation is not ValueTensorLiteralOp");
@@ -169,29 +169,29 @@ public:
     helper.addDenseFloatArrayAttr("scale", {scale});
 
     // zero_point
-    if (!op.getZeroPoint().getDefiningOp())
+    auto zeroPointOp = op.getZeroPoint().getDefiningOp();
+    int64_t zeroPoint;
+    if (!zeroPointOp)
       return rewriter.notifyMatchFailure(op, "Missing zero point operation");
-    if (dyn_cast<torch::Torch::AtenZerosOp>(
-            op.getZeroPoint().getDefiningOp()) ||
-        dyn_cast<torch::Torch::AtenZerosLikeOp>(
-            op.getZeroPoint().getDefiningOp())) {
-      helper.addDenseIntArrayAttr("zero_point", {0});
-      return helper.replace();
+    if (dyn_cast<torch::Torch::AtenZerosOp>(zeroPointOp) ||
+        dyn_cast<torch::Torch::AtenZerosLikeOp>(zeroPointOp)) {
+      zeroPoint = 0;
+    } else {
+      auto zeroPointTensor =
+          dyn_cast<torch::Torch::ValueTensorLiteralOp>(zeroPointOp);
+      if (!zeroPointTensor)
+        return rewriter.notifyMatchFailure(
+            op, "Zero point operation is not ValueTensorLiteralOp or Zero "
+                "operation");
+      auto zeroPointElements =
+          dyn_cast<DenseIntElementsAttr>(zeroPointTensor.getValueAttr());
+      if (!zeroPointElements || zeroPointElements.getNumElements() != 1)
+        return rewriter.notifyMatchFailure(
+            op, "Unsupported zero point type or size");
+      zeroPoint = (*zeroPointElements.begin()).getSExtValue();
     }
-    auto zeroPointTensor = dyn_cast<torch::Torch::ValueTensorLiteralOp>(
-        op.getZeroPoint().getDefiningOp());
-    if (!zeroPointTensor)
-      return rewriter.notifyMatchFailure(
-          op,
-          "Zero point operation is not ValueTensorLiteralOp or Zero operation");
-    auto zeroPointElements =
-        dyn_cast<DenseIntElementsAttr>(zeroPointTensor.getValueAttr());
-    if (!zeroPointElements || zeroPointElements.getNumElements() != 1)
-      return rewriter.notifyMatchFailure(op,
-                                         "Unsupported zero point type or size");
-    auto zeroPoint = (*zeroPointElements.begin()).getSExtValue();
-    helper.addDenseIntArrayAttr("zero_point", {zeroPoint});
 
+    helper.addDenseIntArrayAttr("zero_point", {zeroPoint});
     return helper.replace();
   }
 };
@@ -212,10 +212,10 @@ public:
     helper.addIntAttr("quant_max", op.getQuantMax());
 
     // scale
-    if (!op.getScale().getDefiningOp())
+    auto scaleOp = op.getScale().getDefiningOp();
+    if (!scaleOp)
       return rewriter.notifyMatchFailure(op, "Missing scale operation");
-    auto scaleTensor = dyn_cast<torch::Torch::ValueTensorLiteralOp>(
-        op.getScale().getDefiningOp());
+    auto scaleTensor = dyn_cast<torch::Torch::ValueTensorLiteralOp>(scaleOp);
     if (!scaleTensor)
       return rewriter.notifyMatchFailure(
           op, "Scale operation is not ValueTensorLiteralOp");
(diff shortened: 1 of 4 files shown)


lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -117,6 +117,150 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+    auto scale = (*scaleElements.begin()).convertToDouble();
+    helper.addDenseFloatArrayAttr("scale", {scale});
+
+    // zero_point
+    if (!op.getZeroPoint().getDefiningOp())
+      return rewriter.notifyMatchFailure(op, "Missing zero point operation");
+    if (dyn_cast<torch::Torch::AtenZerosOp>(
+            op.getZeroPoint().getDefiningOp()) ||
+        dyn_cast<torch::Torch::AtenZerosLikeOp>(
+            op.getZeroPoint().getDefiningOp())) {
```
sjain-stanford, line 177: Consider pulling out `op.getZeroPoint().getDefiningOp()` to avoid repetition.

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -117,6 +117,150 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+                                                            getTypeConverter()};
+    helper.addOperand("self", adaptor.getSelf());
+    helper.addIntAttr("quant_min", op.getQuantMin());
+    helper.addIntAttr("quant_max", op.getQuantMax());
+
+    // scale
+    if (!op.getScale().getDefiningOp())
+      return rewriter.notifyMatchFailure(op, "Missing scale operation");
+    auto scaleTensor = dyn_cast<torch::Torch::ValueTensorLiteralOp>(
+        op.getScale().getDefiningOp());
```
sjain-stanford, line 160: Same (`op.getScale().getDefiningOp()` repetition)

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -117,6 +117,150 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+      return rewriter.notifyMatchFailure(
+          op,
+          "Zero point operation is not ValueTensorLiteralOp or Zero operation");
+    auto zeroPointElements =
+        dyn_cast<DenseIntElementsAttr>(zeroPointTensor.getValueAttr());
+    if (!zeroPointElements || zeroPointElements.getNumElements() != 1)
+      return rewriter.notifyMatchFailure(op,
+                                         "Unsupported zero point type or size");
+    auto zeroPoint = (*zeroPointElements.begin()).getSExtValue();
+    helper.addDenseIntArrayAttr("zero_point", {zeroPoint});
```
sjain-stanford, line 193: Consider rewriting this section as follows to make control flow more readable and less error prone to future modifications:
```C++
auto zeroPointOp = op.getZeroPoint().getDefiningOp();

if (dyn_cast<torch::Torch::AtenZerosOp>(zeroPointOp) ||
  dyn_cast<torch::Torch::AtenZerosLikeOp>(zeroPointOp)) {
  auto zeroPoint = -=0;
} else {
  auto zeroPointTensor = ...
  ...
  auto zeroPoint = (*zeroPointElements.begin()).getSExtValue();
}
helper.addDenseIntArrayAttr("zero_point", {zeroPoint});
return helper.replace();
```

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -117,6 +117,150 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+                  ConversionPatternRewriter &rewriter) const override {
+    torch_to_tcp::TorchToTcpCustomOpConversionHelper helper{op, rewriter,
+                                                            getTypeConverter()};
+    helper.addOperand("self", adaptor.getSelf());
+    helper.addIntAttr("axis", op.getQuantMin());
+    helper.addIntAttr("quant_min", op.getQuantMin());
+    helper.addIntAttr("quant_max", op.getQuantMax());
+
+    // scale
+    if (!op.getScale().getDefiningOp())
```
sjain-stanford, line 215: same as above

sjain-stanford (commented):
Mostly looks good, left some suggestions.

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -117,6 +117,147 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+      auto zeroPointTensor =
+          dyn_cast<torch::Torch::ValueTensorLiteralOp>(zeroPointOp);
+      if (!zeroPointTensor)
+        return rewriter.notifyMatchFailure(
+            op, "Zero point operation is not ValueTensorLiteralOp or Zero "
+                "operation");
+      auto zeroPointElements =
+          dyn_cast<DenseIntElementsAttr>(zeroPointTensor.getValueAttr());
+      if (!zeroPointElements ||
+          zeroPointElements.getType().getShape().size() != 1)
```
sjain-stanford, line 249: Curious, is this checking that `zeroPointElements` is a 1D tensor? Please add a comment for clarity.

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -117,6 +117,147 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+public:
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(AtenFakeQuantizePerChannelAffineOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    torch_to_tcp::TorchToTcpCustomOpConversionHelper helper{op, rewriter,
+                                                            getTypeConverter()};
+    helper.addOperand("self", adaptor.getSelf());
+    helper.addIntAttr("axis", op.getQuantMin());
```
sjain-stanford, line 210: Should this be `op.getAxis()`?

test/Conversion/TorchToTcp/tcp_custom_ops.mlir
```diff
@@ -134,3 +134,131 @@ func.func @torch.aten.regular_convolution_2d(%input: !torch.vtensor<[1,9,16,1600
...
+  return %output : !torch.vtensor<[1,64,32,32],f32>
+}
+
+// -----
+
+// CHECK-LABEL: func.func @torch.aten.fake_quantize_per_channel_affine(
+// CHECK-SAME:         %[[ARG0:.*]]: !torch.vtensor<[1,3,32,32],f32>) -> !torch.vtensor<[1,3,32,32],f32>
+// CHECK:          %[[T0:.*]] = torch_c.to_builtin_tensor %[[ARG0]] : !torch.vtensor<[1,3,32,32],f32> -> tensor<1x3x32x32xf32>
+// CHECK:          %[[CUSTOM:.*]] = tcp.custom_op("torch.aten.fake_quantize_per_channel_affine") %[[T0]] {
+// CHECK-SAME:                          axis = 0 : i64,
```
sjain-stanford, line 218: Squinting at it I think `axis` should be 1 here?

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -117,6 +117,147 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+    auto scaleTensor = dyn_cast<torch::Torch::ValueTensorLiteralOp>(scaleOp);
+    if (!scaleTensor)
+      return rewriter.notifyMatchFailure(
+          op, "Scale operation is not ValueTensorLiteralOp");
+    auto scaleElements =
+        dyn_cast<DenseFPElementsAttr>(scaleTensor.getValueAttr());
+    if (!scaleElements || scaleElements.getNumElements() != 1)
+      return rewriter.notifyMatchFailure(op, "Unsupported scale type or size");
+    auto scale = (*scaleElements.begin()).convertToDouble();
+    helper.addDenseFloatArrayAttr("scale", {scale});
```
sjain-stanford, line 169: Just a thought. Scale and zero-point are tensors in the Torch op definition. That means they might be dynamic values. By converting them to attributes, are we making the assumption here that we only support the case when they are static? Or do we want to preserve them as tensors?
```
    AnyTorchTensorType:$scale,
    AnyTorchTensorType:$zero_point,
```
Looping @navahgar on this as I recall him having some opinions on this.

sjain-stanford (commented):
Thanks for addressing. Some more comments and a bugfix in the 2nd pass. If possible, let's wait for @navahgar 's inputs on the scale/zero-point being tensors vs attributes.

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
zezhang: Yes, good catch!

test/Conversion/TorchToTcp/tcp_custom_ops.mlir
zezhang: Fixed by addressing the axis bug.

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
zezhang: there are 2 modes for quantization op: per_tensor and per_channel.

For the per_tensor mode, a single value of scale and zero_point are used to represent the entire tensor. 

- For `AtenFakeQuantizePerTensorAffineOp` , these two values are constant float and int. 
- For the `AtenFakeQuantizePerTensorAffineTensorQparamsOp`, both scale and zero_point operands need to be size-1 ValueTensorLiteral (also constant but in tensor format with shape of [1])

For the per_channel mode, the number of scale and zero_point need to be the channel dimension C. I don't think these can be dynamic since the scale and zero_point need to be compile-time constant.

- For `AtenFakeQuantizePerChannelAffineOp` , both scale and zero_point operands need to be ValueTensorLiteral   constants with shape of [C]

sjain-stanford: > For the per_channel mode, the number of scale and zero_point need to be the channel dimension C. I don't think these can be dynamic since the scale and zero_point need to be compile-time constant.

Per https://pytorch.org/docs/stable/generated/torch.fake_quantize_per_channel_affine.html it appears to allow any tensors for scale and zero-point, so they could be dynamic valued (shape is fixed by channel/feature length).

sjain-stanford: I think @navahgar had some thoughts on this, just don't recall off the top of my head. Since he's on vacation, I'm fine with landing this with documentation, and circling back if this needs to be changed later.

sjain-stanford (approved):

PR #81, opened 2024-07-15
[AOT] Use torch dialect in backend form instead of raw torch dialect

sjain-stanford:
With https://github.com/llvm/torch-mlir/pull/3541 we can now intercept the Torch dialect during TorchDynamo export at two stages:

1. OutputType.RAW: This gets us the torch dialect as-imported from the FX graph
2. OutputType.TORCH: This gets us the torch dialect in backend compliant form, after the raw torch goes through DecomposeComplexOps and ReduceOpVariants.

We've been using 1 for all the AOT (e2e) tests, however this PR changes it to use 2, which is closer to the real backend lowering pipelines we use internally as well.

### DIFF
--- a/requirements_lock.txt
+++ b/requirements_lock.txt
@@ -6,15 +6,15 @@
 #
 --find-links https://github.com/llvm/torch-mlir-release/releases/expanded_assets/dev-wheels
 
-filelock==3.15.3 \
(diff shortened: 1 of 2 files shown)


zezhang (approved):
LGTM!

PR #82, opened 2024-07-15
Handle `tcp.bind_symbolic_shape` ops in fusion algorithm

srinathava:
The current fusion algorithm does not handle `tcp.bind_symbolic_shape` ops very well because these ops cause an op to have multiple downstream uses. This change handles these ops "specially". We collect all the uses of an op which only specify the shape of the output into a separate use category. We then move the bind shape ops along with the original op if it was moved to another region.

### DIFF
--- a/lib/Dialect/Transforms/FusionPatterns.cpp
+++ b/lib/Dialect/Transforms/FusionPatterns.cpp
@@ -65,7 +65,7 @@ GenericBottomUpFuser::matchAndRewrite(Operation *op,
       uses.push_back(use.getOwner());
   }
 
-  // All its uses are tcp.bind_symbolic_shpae ops.
+  // All its uses are tcp.bind_symbolic_shape ops.
   if (uses.empty())
     return failure();
 
--- a/lib/Dialect/Transforms/FusionPatterns.cpp
+++ b/lib/Dialect/Transforms/FusionPatterns.cpp
@@ -44,8 +44,15 @@ GenericBottomUpFuser::matchAndRewrite(Operation *op,
   Region *usesParentRegion = nullptr;
   SmallVector<Operation *> uses;
   llvm::DenseSet<Operation *> usesSet;
+  llvm::DenseSet<tcp::BindSymbolicShapeOp> bindShapeUses;
 
+  LLVM_DEBUG(llvm::dbgs() << "Processing op: " << *op << "\n");
   for (auto &use : op->getUses()) {
+    if (auto bindShapeOp = dyn_cast<tcp::BindSymbolicShapeOp>(use.getOwner())) {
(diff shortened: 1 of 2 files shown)


lib/Dialect/Transforms/FusionPatterns.cpp
```diff
@@ -58,6 +65,10 @@ GenericBottomUpFuser::matchAndRewrite(Operation *op,
       uses.push_back(use.getOwner());
   }
 
+  // All its uses are tcp.bind_symbolic_shpae ops.
```
sjain-stanford, line 68: nit: typo

sjain-stanford (commented):
Could you please add some lit tests?

test/Dialect/tcp_fusion.mlir
```diff
@@ -97,3 +97,38 @@ func.func @test_multi_use_fusion(%arg0 : tensor<?x?xf32>, %arg1 : tensor<?x?xf32
...
+// CHECK:     tcp.bind_symbolic_shape %[[ARG0]], [%[[V0]], %[[V1]]], affine_map<()[s0, s1] -> (s0, s1)> : tensor<?x?xf32>
+// CHECK:     %[[V2:.+]] = tcp.group {
+// CHECK:       %[[V3:.+]] = tcp.tanh %[[ARG0]] : tensor<?x?xf32> -> tensor<?x?xf32>
+// CHECK:       tcp.bind_symbolic_shape %[[V3]], [%[[V0]], %[[V1]]], affine_map<()[s0, s1] -> (s0, s1)> : tensor<?x?xf32>
+// CHECK:       %[[V4:.+]] = tcp.tanh %[[V3]] : tensor<?x?xf32> -> tensor<?x?xf32>
+// CHECK:       tcp.bind_symbolic_shape %[[V4]], [%[[V0]], %[[V1]]], affine_map<()[s0, s1] -> (s0, s1)> : tensor<?x?xf32>
+// CHECK:       %[[V5:.+]] = tcp.add %[[V4]], %[[V4]] : tensor<?x?xf32>, tensor<?x?xf32> -> tensor<?x?xf32>
+// CHECK:       tcp.yield %[[V5]] : tensor<?x?xf32>
+// CHECK:     } : tensor<?x?xf32>
+// CHECK:     tcp.bind_symbolic_shape %[[V2]], [%[[V0]], %[[V1]]], affine_map<()[s0, s1] -> (s0, s1)> : tensor<?x?xf32>
```
sjain-stanford, line 115: Does it work the same way when the tcp.group returns more than one output? i.e. the `tcp.bind_symbolic_shape` ops are outside the groups?

sjain-stanford (approved):
Thanks! LGTM besides one question.

PR #83, opened 2024-07-16
update checking logic for scale and zp op

zezhang:
For fake_quantize_ops, `scale` and `zero_point` operands could be lowered by `TorchToTcp` already, so we need to relax the checking condition here (just need to make sure the type and shape are valid).


sjain-stanford (approved):
LGTM with a comment request.

PR #85, opened 2024-07-24
Revert multi-use fusion algorithm

srinathava:
Revert the more aggressive multi-use fusion algorithm given that it has a correctness bug.

The newly added lit test demonstrates the correctness issue. 

The only part of the new algorithm which is retained is the special handling of `tcp.bind_symbolic_shape` ops.


navahgar (approved):
LGTM

PR #87, opened 2024-07-30
Add support for customizing const inlining and ops with regions to IsolateGroupOps pass

srinathava:
A few small improvements to the `IsolateGroupOps` pass:
- Provide a hook to customize whether a const like op used in the `tcp.group` will get copied into the group or whether it will be passed in as an input argument. The main pass does always returns true so this is a non-functional change in `mlir-tcp`. 
- The previous version did not handle ops with contained regions and block arguments (such as an `scf.forall` inside the `tcp.group`). We now handle this (see updated lit test).

### DIFF
--- a/lib/Dialect/Transforms/IsolateGroupOpsPass.cpp
+++ b/lib/Dialect/Transforms/IsolateGroupOpsPass.cpp
@@ -105,6 +105,20 @@ public:
   std::function<bool(tcp::GroupOp, Value)> shouldInlineConst_;
 };
 
+class DropSymbolicShapesInsideGroups
+    : public OpRewritePattern<BindSymbolicShapeOp> {
+  using OpRewritePattern<BindSymbolicShapeOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(BindSymbolicShapeOp shapeOp,
+                                PatternRewriter &rewriter) const override {
+    if (isa<tcp::GroupOp>(shapeOp->getParentOp())) {
+      rewriter.eraseOp(shapeOp);
+      return success();
+    }
+    return failure();
+  }
+};
+
 class TcpIsolateGroupOpsPass
     : public TcpIsolateGroupOpsBase<TcpIsolateGroupOpsPass> {
   void runOnOperation() override {
@@ -130,6 +144,7 @@ void populateIsolateGroupPatterns(
     std::function<bool(tcp::GroupOp, Value)> shouldCopyConstPredicate) {
 
   patterns.add<IsolateGroups>(patterns.getContext(), shouldCopyConstPredicate);
+  patterns.add<DropSymbolicShapesInsideGroups>(patterns.getContext());
 }
 
 } // namespace mlir::tcp
--- a/lib/Dialect/Transforms/IsolateGroupOpsPass.cpp
+++ b/lib/Dialect/Transforms/IsolateGroupOpsPass.cpp
@@ -29,14 +29,16 @@ namespace mlir::tcp {
 
 namespace {
 
-class IsolateGroups : public OpRewritePattern<tcp::GroupOp> {
+class IsolateGroups : public OpRewritePattern<GroupOp> {
 public:
-  IsolateGroups(MLIRContext *ctx,
-                std::function<bool(Operation *)> shouldCopyConstPredicate)
-      : OpRewritePattern(ctx),
-        _shouldCopyConstPredicate(shouldCopyConstPredicate) {}
+  using OpRewritePattern<GroupOp>::OpRewritePattern;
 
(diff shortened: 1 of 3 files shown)


lib/Dialect/Transforms/IsolateGroupOpsPass.cpp
```diff
@@ -100,7 +112,8 @@ class TcpIsolateGroupOpsPass
     MLIRContext *context = op->getContext();
     RewritePatternSet patterns(context);
 
-    patterns.add<IsolateGroups>(context);
+    auto shouldCopyConstPredicate = [&](tcp::GroupOp, Value) { return true; };
+    populateIsolateGroupPatterns(patterns, shouldCopyConstPredicate);
```
navahgar, line 116: Couple of questions:
1. Whats the adv of having a lambda over a bool to control this?
2. Why are we passing this to `populateIsolateGroupPatterns` instead of adding it as an option to the pass itself?

srinathava: Recapping offline discussion:
1. The downstream uses of this API do have more complicated logic to figure out when to inline a const (for instance based on users of that const, compilation mode etc.)
2. I am not sure there's an easy way to add a boolean predicate function as an option to a pass itself in the .td file. I'd be happy to also add a default functor argument to the pass in that case.

navahgar (approved):
LGTM

lib/Dialect/Transforms/IsolateGroupOpsPass.cpp
```diff
@@ -29,45 +29,55 @@ namespace mlir::tcp {
 
 namespace {
 
-class IsolateGroups : public OpRewritePattern<tcp::GroupOp> {
+class IsolateGroups : public OpRewritePattern<GroupOp> {
 public:
-  using OpRewritePattern::OpRewritePattern;
+  using OpRewritePattern<GroupOp>::OpRewritePattern;
```
sjain-stanford, line 34: nit: I think we can drop this change

lib/Dialect/Transforms/IsolateGroupOpsPass.cpp
```diff
@@ -91,6 +101,8 @@ class IsolateGroups : public OpRewritePattern<tcp::GroupOp> {
     rewriter.eraseOp(groupOp);
     return success();
   }
+
+  std::function<bool(tcp::GroupOp, Value)> shouldInlineConst_;
```
sjain-stanford, line 105: nit: consider making this a private member or removing the trailing underscore.

lib/Dialect/Transforms/IsolateGroupOpsPass.cpp
sjain-stanford, line 116: Just for my understanding, is the idea here that in the downstream (internal) use we'd define a separate pass that calls `populateIsolateGroupPatterns` with a custom `shouldCopyConstPredicate` functor? Was there an alternative that also avoids the duplication of the pass definition itself?

lib/Dialect/Transforms/IsolateGroupOpsPass.cpp
```diff
@@ -29,45 +29,55 @@ namespace mlir::tcp {
...
             addedInputs.insert(operand);
           }
         }
       }
-      defs.insert(op.getResults().begin(), op.getResults().end());
-    }
+    });
 
-    auto isolatedGroupOp = rewriter.create<tcp::IsolatedGroupOp>(
+    auto isolatedGroupOp = rewriter.create<IsolatedGroupOp>(
```
sjain-stanford, line 69: minor nit (feel free to ignore): consider keeping the `tcp::` scope to be consistent with other transforms.

lib/Dialect/Transforms/IsolateGroupOpsPass.cpp
```diff
@@ -29,45 +29,55 @@ namespace mlir::tcp {
 
 namespace {
 
-class IsolateGroups : public OpRewritePattern<tcp::GroupOp> {
+class IsolateGroups : public OpRewritePattern<GroupOp> {
```
sjain-stanford, line 32: same

sjain-stanford (approved):
Thanks for this tech-debt cleanup @srinathava ! Besides a few requests, LGTM.

lib/Dialect/Transforms/IsolateGroupOpsPass.cpp
srinathava: Yes. that would be the idea. Looking around (admittedly not very hard), I did not find a way to add a predicate function as an argument to a pass constructor. Admittedly the downstream pass will be a bit bare...

PR #89, opened 2024-08-14
fix broadcast condition

zezhang:
As titled.

### DIFF
--- a/lib/Conversion/TorchToTcp/Misc.cpp
+++ b/lib/Conversion/TorchToTcp/Misc.cpp
@@ -82,6 +82,8 @@ public:
 
     SmallVector<int64_t> axes;
     SmallVector<Value> resultShape;
+    ArrayRef<int64_t> newInputShape =
+        input.getType().dyn_cast<RankedTensorType>().getShape();
     for (int64_t i = 0; i < static_cast<int64_t>(newDimSizes.size()); ++i) {
       Value newDimSize = newDimSizes[i];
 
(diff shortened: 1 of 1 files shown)


navahgar (commented):
Can you add some tests for this, both valid and invalid cases? 

Also, what about the verifier? Does that have to be updated as well?

lib/Conversion/TorchToTcp/Misc.cpp
```diff
@@ -82,6 +82,8 @@ class ConvertAtenBroadcastLikeOps : public OpConversionPattern<AtenOpT> {
 
     SmallVector<int64_t> axes;
     SmallVector<Value> resultShape;
+    ArrayRef<int64_t> newInputShape =
+        input.getType().dyn_cast<RankedTensorType>().getShape();
```
AaronStGeorge, line 86: This seems the same a the variable `inputShape` that already exists?

lib/Conversion/TorchToTcp/Misc.cpp
```diff
@@ -114,11 +116,14 @@ class ConvertAtenBroadcastLikeOps : public OpConversionPattern<AtenOpT> {
       // Note: The order of checks in this boolean expression matters!
       if (isNewDim || isDynamicDim ||
           (!isDimSizePreserved && doesDimSizeChange)) {
-        axes.push_back(i);
-        newDimSize = rewriter.create<torch::TorchConversion::ToI64Op>(
-            op->getLoc(), newDimSize);
-        resultShape.push_back(rewriter.create<arith::IndexCastOp>(
-            op->getLoc(), rewriter.getIndexType(), newDimSize));
+        // only broadcast if the input dim size is 1
+        if (newInputShape[i] == 1) {
```
AaronStGeorge, line 120: This might be a little nicer with two booleans something like
```diff
+bool inputDimBroadcastable = newInputShape[i] == 1);
+// Note: The order of checks in this boolean expression matters!
+bool outputDimBroadcasted = isNewDim || isDynamicDim ||
+          (!isDimSizePreserved && doesDimSizeChange);
-// Note: The order of checks in this boolean expression matters!
-if (isNewDim || isDynamicDim ||
-          (!isDimSizePreserved && doesDimSizeChange)) {
+if (outputDimBroadcasted && inputDimBroadcastable) {

zezhang:
> Can you add some tests for this, both valid and invalid cases?
> 
> Also, what about the verifier? Does that have to be updated as well?

No need to update the verifier as it checks all axes that need to be broadcasted must have static size of 1

lib/Conversion/TorchToTcp/Misc.cpp
zezhang: `inputShape` is before the `broadcastRankInLeadingDims`. We will need to use the `newInputShape`, which is after the rank expansion (if applicable) to match the rank of output shape

lib/Conversion/TorchToTcp/Misc.cpp
zezhang: Yup I like this idea. Will update it.

zezhang:
Updated in : https://github.com/cruise-automation/mlir-tcp/pull/90

PR #90, opened 2024-08-14
fix broadcast condition

zezhang:
As titled.

To test:

`bazel test //...`

### DIFF
--- a/lib/Conversion/TorchToTcp/Misc.cpp
+++ b/lib/Conversion/TorchToTcp/Misc.cpp
@@ -113,17 +113,16 @@ public:
               ? true
               : staticDimSize != inputShape[i - newLeadingDims];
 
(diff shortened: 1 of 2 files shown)


AaronStGeorge (approved):
LGTM

lib/Conversion/TorchToTcp/Misc.cpp
```diff
@@ -111,9 +113,11 @@ class ConvertAtenBroadcastLikeOps : public OpConversionPattern<AtenOpT> {
               ? true
               : staticDimSize != inputShape[i - newLeadingDims];
 
+      bool inputDimBroadcastable = newInputShape[i] == 1;
```
AaronStGeorge, line 116: Not a big deal but you could prefix `is`

PR #92, opened 2024-08-30
update custom op conversions

zezhang:
- Fix a minor issue with broadcast op (fold the broadcast if axes attr is empty)
- Add following ops to tcp.custom_op:
--`aten.sort`
--`aten.cumsum`
--`aten.min.dim`
--`aten.view`(dynamic shape only)
--`aten.topk`

To test:
`bazel test //...` (in docker)

### DIFF
--- a/lib/Conversion/TorchToTcp/TcpCustomOp.cpp
+++ b/lib/Conversion/TorchToTcp/TcpCustomOp.cpp
@@ -14,9 +14,12 @@
 
 #include "PopulatePatterns.h"
 #include "Utils.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "torch-mlir/Conversion/Utils/Utils.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
+#include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionOps.h"
 
 #include "llvm/ADT/StringSet.h"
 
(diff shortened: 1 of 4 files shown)


lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -211,6 +214,118 @@ class ConvertAtenFakeQuantizePerChannelAffineOp
...
+                                                            getTypeConverter()};
+    Value self = adaptor.getSelf();
+    auto srcType = self.getType().cast<RankedTensorType>();
+    auto resultType =
+        getTypeConverter()->convertType(op.getType()).cast<RankedTensorType>();
+
+    SmallVector<int64_t> size;
+    if (matchPattern(op.getSize(), m_TorchListOfConstantInts(size)) &&
+        srcType.hasStaticShape() && resultType.hasStaticShape())
+      return rewriter.notifyMatchFailure(op, "only dynamic shape is supported");
```
navahgar, line 289: Can you add a comment here to say why this is the case? I understand the reason but it would be good to document it.

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -211,6 +214,118 @@ class ConvertAtenFakeQuantizePerChannelAffineOp
...
+    Operation *primListOp = op.getSize().getDefiningOp();
+    auto listConstruct = dyn_cast<Torch::PrimListConstructOp>(primListOp);
+    if (!listConstruct) {
+      return rewriter.notifyMatchFailure(
+          op, "Size must come from PrimListConstructOp");
+    }
+    int idx = 0;
+    for (Value value : listConstruct.getElements()) {
+      int64_t dimSize;
+      if (!matchPattern(value, m_TorchConstantInt(&dimSize))) {
```
navahgar, line 301: nit: given that you handle `else` case below it might be clearer to negate this condition. Just makes things easier to read.

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -211,6 +214,118 @@ class ConvertAtenFakeQuantizePerChannelAffineOp
...
+          return rewriter.notifyMatchFailure(
+              op, "dynamic dim size should come from IndexCastOp");
+        auto indexCastOp = dyn_cast<arith::IndexCastOp>(
+            conversionOp.getOperand().getDefiningOp());
+        if (!isa<tensor::DimOp>(indexCastOp.getIn().getDefiningOp()))
+          return rewriter.notifyMatchFailure(
+              op, "dynamic dim size should come from DimOp");
+        auto dimOp =
+            dyn_cast<tensor::DimOp>(indexCastOp.getIn().getDefiningOp());
+        helper.addOperand("idx_" + std::to_string(idx), dimOp);
```
navahgar, line 318: I see you are matching a specific pattern. Can you add a comment to show the pattern? That would make it easier to read the code.

navahgar (approved):
LGTM. Some minor comments inline.

lib/Conversion/TorchToTcp/TcpCustomOp.cpp
```diff
@@ -211,6 +214,118 @@ class ConvertAtenFakeQuantizePerChannelAffineOp
   }
 };
 
+class ConvertAtenSortOp : public OpConversionPattern<AtenSortOp> {
```
matthewfl, line 217: Most likely the sort is coming from a top-k.  I would prefer that we keep top-k instead of changing it to a sort as trt supports top-k already, and if we add sort then it just ends up getting changed back to top-k internally with additional performance penalties due to any extra ops that get added

zezhang: Good idea. We can add an internal patch [here](https://sourcegraph.robot.car/github.robot.car/cruise/mla-robocomp-torch-mlir/-/blob/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp?L8848) to disable the topk-->sort. I will add a new custom_op to handle the topk then.

PR #94, opened 2024-09-05
relax aten.view conversion constraint 

zezhang:
Only convert aten.view to tcp.custom_op when the `size` array is non-constant. The rest will be handled through torch-to-tosa.

TODO: send out an upstream PR to fix` tosa.reshape`  size calculation logic.


navahgar (approved):
LGTM

PR #95, opened 2024-09-05
add scatter to tcp custom op

matthewfl:
as titled


sjain-stanford (commented):
Please add a lit test.

PR #98, opened 2024-09-18
Add converter for index.Tensor_hacked_twin

matthewfl:
* Add converter for index.Tensor_hacked_twin -> tcp.gather & tcp.broadcast
* Remove the  `tcp.custom_op` variants of `Tensor_hacked_twin`
  * The Tcp variants do not have full coverage of the PyTorch, but we should seek to expand the coverage of our converters
* Fix tcp.const when it is a dense resource of ints to cast the value
* Add verifier for `tcp.gather` and `tcp.const` to ensure that is used correctly

### DIFF
--- a/lib/Dialect/IR/TcpOps.cpp
+++ b/lib/Dialect/IR/TcpOps.cpp
@@ -170,6 +170,27 @@ LogicalResult CastOp::verify() {
   return success();
 }
 
+LogicalResult GatherOp::verify() {
+  auto inputTensor = cast<RankedTensorType>(getInput().getType());
+    auto indicesTensor = cast<RankedTensorType>(getIndices().getType());
+    int64_t gatherDim = getDimAttr().getValue().getSExtValue();
+
+   if(inputTensor.getRank() != indicesTensor.getRank())
+      return emitOpError("tcp.gather requires that the input tensor and indices are the same rank");
+
+    for(int i = 0; i < inputTensor.getRank(); i++) {
+      if(inputTensor.getShape()[i] != indicesTensor.getShape()[i]) {
+        if(!(inputTensor.getShape()[i] == ShapedType::kDynamic ||
+             indicesTensor.getShape()[i] == 1 ||
+             i == gatherDim)) {
+              return emitOpError("indices tensor does not match expected shape");
+             }
+      }
+    }
+
+    return success();
+}
+
 //===----------------------------------------------------------------------===//
 // BindSymbolicShapeOp
 //===----------------------------------------------------------------------===//
--- a/lib/Dialect/IR/TcpOps.cpp
+++ b/lib/Dialect/IR/TcpOps.cpp
@@ -179,13 +179,15 @@ LogicalResult GatherOp::verify() {
       return emitOpError("tcp.gather requires that the input tensor and indices are the same rank");
 
     for(int i = 0; i < inputTensor.getRank(); i++) {
-      if(inputTensor.getShape()[i] != indicesTensor.getShape()[i]) {
-        if(!(inputTensor.getShape()[i] == ShapedType::kDynamic ||
-             indicesTensor.getShape()[i] == 1 ||
+      if(inputTensor.getShape()[i] != indicesTensor.getShape()[i] && !(
+        inputTensor.getShape()[i] == ShapedType::kDynamic ||
              i == gatherDim)) {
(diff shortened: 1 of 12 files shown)


test/Conversion/TorchToTcp/data_movement.mlir
```diff
@@ -64,3 +64,29 @@ func.func @torch.aten.index_select(%arg0: !torch.vtensor<[4,3],f32>, %arg1: !tor
...
+// CHECK: return %[[RET]]
+func.func @torch.aten.index.tensor_hacked_twin(%arg0: !torch.vtensor<[1,20,30],f32>, %select1: !torch.vtensor<[5,1],si64>, %select2: !torch.vtensor<[20],si64>) -> !torch.vtensor<[1,5,20],f32> {
+  // there is a strange pattern that is being generated when selecting one axis.  It seems that it uses the Tensor_hacked_twin to select along all axis, but uses
+  // arange to select all of the 
+  %none = torch.constant.none
+  %int0 = torch.constant.int 0
+  %int1 = torch.constant.int 1
+  %int4 = torch.constant.int 4 // this is a dtype on arange....
+  %int-1 = torch.constant.int -1
+  %arange = torch.aten.arange.start_step %int0, %int1, %int1, %int4, %none, %none, %none : !torch.int, !torch.int, !torch.int, !torch.int, !torch.none, !torch.none, !torch.none -> !torch.vtensor<[1],si64>
```
zezhang, line 85: `torch.aten.arange.start_step` will remain unchanged right? As we don't handle it in tcp.

zezhang (commented):
LGTM! Can we also try to add it to the [aot_compile](https://github.com/cruise-automation/mlir-tcp/blob/main/test/AotCompile/model_loader_lib.py) test suite for e2e coverage?

test/Conversion/TorchToTcp/data_movement.mlir
matthewfl: Currently this remains unchanged.  I was thinking that we should add an optimization pass that removes this when it selects everything, but I have yet to do that

lib/Dialect/IR/TcpOps.cpp
```diff
@@ -170,6 +176,31 @@ LogicalResult CastOp::verify() {
...
+    return emitOpError("tcp.gather requires that the input tensor and indices "
+                       "are the same rank");
+
+  for (int i = 0; i < inputTensor.getRank(); i++) {
+    if (inputTensor.getShape()[i] != indicesTensor.getShape()[i] &&
+        !(inputTensor.getShape()[i] == ShapedType::kDynamic ||
+          i == gatherDim)) {
+      return emitOpError("indices tensor does not match expected shape");
+    }
+  }
```
navahgar, line 194: This doesn't seem to match the semantics of `tcp.gather`, which should have the same constraints as `torch.gather`.

Specifically the following from [here](https://pytorch.org/docs/stable/generated/torch.gather.html#torch.gather)
> It is also required that index.size(d) <= input.size(d) for all dimensions d != dim

matthewfl: I updated it

lib/Dialect/IR/TcpOps.cpp
```diff
@@ -170,6 +176,32 @@ LogicalResult CastOp::verify() {
...
+  if (inputTensor.getRank() != indicesTensor.getRank())
+    return emitOpError("tcp.gather requires that the input tensor and indices "
+                       "are the same rank");
+
+  for (int i = 0; i < inputTensor.getRank(); i++) {
+    if (inputTensor.getShape()[i] < indicesTensor.getShape()[i] &&
+        !(inputTensor.getShape()[i] == ShapedType::kDynamic ||
+          indicesTensor.getShape()[i] == ShapedType::kDynamic ||
+          i == gatherDim)) {
+      return emitOpError("indices tensor does not match expected shape");
```
navahgar, line 193: Can you update this error to explicitly say the exact condition that fails here? That might be useful to know.

navahgar (approved):
LGTM for `GatherOp::verify()`

lib/Dialect/IR/TcpOps.cpp
matthewfl: done

zezhang (approved):
LGTM! Thanks for working on this tricky case.

PR #99, opened 2024-10-02
[NFC] Bump torch-mlir and torch pip deps

sjain-stanford:
This was long overdue, but we had torch-mlir releases [broken](https://github.com/llvm/torch-mlir-release/actions/runs/10838435405) for a while in September, and the new ones were released under a different name (fixed by https://github.com/llvm/torch-mlir/pull/3752).

Also, since torch-mlir is no longer released with PT extensions, I had to explicitly specify the cpu nightly link in requirements, otherwise it pulls the gpu torch with a bunch of unnecessary nvidia/cuda deps.

Bump `torch` from `2.4.0.dev20240604+cpu` to `2.6.0.dev20241002+cpu`.
Bump `torch-mlir` from `20240714.152` to `20241002.240`.


navahgar (approved):
LGTM

PR #100, opened 2024-10-07
tcp.custom_op support for torch.aten.arange with dynamic input

zezhang:
tcp.custom_op support for` torch.aten.arange` with dynamic input. Static case will be handled through TOSA dialect.

To test:

(in docker)
`bazel test //...`

### DIFF
--- a/lib/Conversion/TorchToTcp/Utils.cpp
+++ b/lib/Conversion/TorchToTcp/Utils.cpp
@@ -529,6 +529,29 @@ void TorchToTcpCustomOpConversionHelper::addFloatAttr(std::string attrName,
       rewriter.getNamedAttr(attrName, rewriter.getF64FloatAttr(constVal)));
 }
 
+bool TorchToTcpCustomOpConversionHelper::tryConvertConstToFloatAttr(
+    std::string attrName, Value value) {
+  if (conversionResult.failed())
+    return false;
+
+  double constFPVal;
+  if (matchPattern(value, torch::Torch::m_TorchConstantFloat(&constFPVal))) {
+    attrs.push_back(
+        rewriter.getNamedAttr(attrName, rewriter.getF64FloatAttr(constFPVal)));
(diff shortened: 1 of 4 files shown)


lib/Conversion/TorchToTcp/Utils.cpp
```diff
@@ -529,6 +529,29 @@ void TorchToTcpCustomOpConversionHelper::addFloatAttr(std::string attrName,
...
+    return true;
+  }
+
+  // convert constant int to fp if possible
+  int64_t constIntVal;
+  if (matchPattern(value, torch::Torch::m_TorchConstantInt(&constIntVal))) {
+    attrs.push_back(rewriter.getNamedAttr(
+        attrName, rewriter.getF64FloatAttr(static_cast<double>(constIntVal))));
+    return true;
+  }
```
navahgar, line 550: Why do we have to convert int to fp here? Why not convert them to `IntAttr`?

navahgar (commented):
Mostly LGTM. One minor comment inline.

lib/Conversion/TorchToTcp/Utils.cpp
zezhang: `aten.arange.start_step` can take either `int` or `fp` for start and step arguments. However, trt-mlir only accepts [F64Attr](https://sourcegraph.robot.car/github.robot.car/cruise/mla-robocomp-tensorrt-mlir/-/blob/include/mlir-tensorrt/Dialect/TensorRT/IR/TensorRTOps.td?L786-787). Converting all constants to FPAttr saves us some time for type checking.

I can also add the conversion in tcp-to-trtmlir. LMKWYT

navahgar: Thanks for the clarification. Can you add a comment in the code to explain this?

navahgar (approved):

PR #101, opened 2024-10-18
Add tcp.gather_nd and rework index.Tensor_hacked_twin to use gather_nd

matthewfl:
Add tcp.gather_nd and rework index.Tensor_hacked_twin to use gather_nd

### DIFF
--- a/include/mlir-tcp/Dialect/IR/TcpOps.td
+++ b/include/mlir-tcp/Dialect/IR/TcpOps.td
@@ -664,7 +664,7 @@ def Tcp_GatherOp : Tcp_Op<"gather", [Pure, AllElementTypesMatch<["input", "out"]
 
 def Tcp_GatherNDOp : Tcp_Op<"gather_nd", [Pure, AllElementTypesMatch<["input", "out"]>]> {
 
-  let summary = "Gather elements from input based on indices over numtiple dimentions";
+  let summary = "Gather elements from input based on indices over multiple dimentions";
 
   let description = [{
     Gathers elements from a given tensor based on indices that index along multiple dimensions.
--- a/include/mlir-tcp/Dialect/IR/TcpOps.td
+++ b/include/mlir-tcp/Dialect/IR/TcpOps.td
@@ -662,6 +662,30 @@ def Tcp_GatherOp : Tcp_Op<"gather", [Pure, AllElementTypesMatch<["input", "out"]
   let hasVerifier = 1;
 }
 
+def Tcp_GatherNDOp : Tcp_Op<"gather_nd", [Pure, AllElementTypesMatch<["input", "out"]>]> {
+
+  let summary = "Gather elements from input based on indices over numtiple dimentions";
+
+  let description = [{
+    Gathers elements from a given tensor based on indices that index along multiple dimensions.
+
+    More details regarding this op: docs/gather.md
+  }];
+
+  let arguments = (ins
+    Tcp_Tensor:$input,
+    Tcp_IntTensor:$indices
+  );
+
+  let results = (outs
+    Tcp_Tensor:$out
+  );
+
+  let assemblyFormat = "$input `,` $indices attr-dict `:` type($input) `,` type($indices) `->` type($out)";
+
+  let hasVerifier = 1;
+}
+
 def Tcp_SliceOp : Tcp_Op<"slice", [Pure, AllElementTypesMatch<["in", "out"]>, SameVariadicOperandSize]> {
 
   let summary = "Extracts a slice of the input tensor";
--- a/include/mlir-tcp/Dialect/IR/TcpOps.td
+++ b/include/mlir-tcp/Dialect/IR/TcpOps.td
@@ -664,7 +664,7 @@ def Tcp_GatherOp : Tcp_Op<"gather", [Pure, AllElementTypesMatch<["input", "out"]
 
 def Tcp_GatherNDOp : Tcp_Op<"gather_nd", [Pure, AllElementTypesMatch<["input", "out"]>]> {
 
-  let summary = "Gather elements from input based on indices over multiple dimentions";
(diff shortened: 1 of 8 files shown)


lib/Conversion/TorchToTcp/Utils.cpp
```diff
@@ -130,6 +156,98 @@ Value broadcastShapeExceptDims(ConversionPatternRewriter &rewriter, Value input,
...
+    auto type = cast<RankedTensorType>(v.getType());
+    v = broadcastRankInLeadingDims(rewriter, v, maxRank - type.getRank());
+    ret.push_back(v);
+  }
+
+  // figure out what the shape should be for each dim
+  struct ShapeInfo {
+    Value value;
+    bool found = false;
+    int64_t static_value = 1;
```
navahgar, line 186: nit: this variable is not camel case, while the rest of the code follows that.

lib/Conversion/TorchToTcp/Utils.cpp
```diff
@@ -130,6 +156,98 @@ Value broadcastShapeExceptDims(ConversionPatternRewriter &rewriter, Value input,
...
+  }
+
+  for (auto v : values) {
+    auto type = cast<RankedTensorType>(v.getType());
+    v = broadcastRankInLeadingDims(rewriter, v, maxRank - type.getRank());
+    ret.push_back(v);
+  }
+
+  // figure out what the shape should be for each dim
+  struct ShapeInfo {
```
navahgar, line 183: IIUC, this is referring to a specific dim, right? 

We generally use "shape" to refer to the full shape, which is a combination of all dims. Can we rename this to something else more representative, maybe `DimInfo`?

lib/Conversion/TorchToTcp/Utils.cpp
```diff
@@ -130,6 +156,98 @@ Value broadcastShapeExceptDims(ConversionPatternRewriter &rewriter, Value input,
...
+    Value value;
+    bool found = false;
+    int64_t static_value = 1;
+  };
+  SmallVector<ShapeInfo> shapes(maxRank);
```
navahgar, line 188: Is this `shapes` variable actually referring to a single shape?

navahgar (approved):
LGTM. Thanks for taking this up.

Some minor comments inline.

lib/Conversion/TorchToTcp/Utils.cpp
matthewfl: All of the input tensors need to be broadcast to the same shape, so this code is going through all of the tensors to try and determine what should be the resulting shape.  A dimension having a value of 1 means that it can be broadcast.  However, we can also find ourselves in the case where there are more than one tensor that have a non-1 value for the same dimension, in which case we should check that the two values are the same.

E.g. if you had tensors `1x5x2` and `3x1x2` then that would broadcast both of them to be `3x5x2`

navahgar: So, this `shapes` should refer to the final shape to which all the input tensors should be broadcasted to, right? In the example above, it must be `3x5x2` right?

My confusion is from whether this must be one shape or multiple shapes as the name `shapes` is referring to.

sjain-stanford:
Curious, does this functionality need additional e2e tests in `aot_compile` or are existing tests already cover this op? If former, could you include e2e tests as well?

lib/Conversion/TcpToLinalg/DataMovement.cpp
```diff
@@ -91,6 +91,101 @@ class ConvertGatherOp : public OpConversionPattern<GatherOp> {
   }
 };
 
+/**
+ * tcp.gather_nd is lowered to linalg.generic, which allows us to define every
+ * element in the result tensor using a programmatic expression.  The last
+ * dimension of the indicies tensor is used to index into the input tensor.
+ *
+ * For example, we we have an indices tensor of shape 9x4x3x2 and an input
```
sjain-stanford, line 99: nit: "we" appears twice

lib/Conversion/TorchToTcp/DataMovement.cpp
```diff
@@ -278,75 +278,63 @@ class ConvertAtenIndexSelectOp : public OpConversionPattern<AtenIndexSelectOp> {
   }
 };
 
+/**
+ * The index.Tensor_hacked_twin takes a list of tensors which have to be
+ * broadcast together to be the same shape, and then those are feed into a
```
sjain-stanford, line 283: nit: fed

include/mlir-tcp/Dialect/IR/TcpOps.td
```diff
@@ -662,6 +662,30 @@ def Tcp_GatherOp : Tcp_Op<"gather", [Pure, AllElementTypesMatch<["input", "out"]
   let hasVerifier = 1;
 }
 
+def Tcp_GatherNDOp : Tcp_Op<"gather_nd", [Pure, AllElementTypesMatch<["input", "out"]>]> {
+
+  let summary = "Gather elements from input based on indices over multiple dimentions";
```
sjain-stanford, line 667: nit: dimensions

matthewfl:
@sjain-stanford there is an e2e test for the `index.Tensor_hacked_twin` already.  This PR changes the `index.Tensor_hacked_twin` to use `tcp.gather_nd` whereas before it was using `tcp.gather`.  The reason for the change is that more complex indexing that is supported by `index.Tensor_hacked_twin` requires being able to gather over multiple dimensions at the same time.  E.g. you can use `index.Tensor_hacked_twin` to select the diagonal of a matrix by doing something like `x[torch.arange(x.shape[0]), torch.arange(x.shape[0])]`.

I will fix the nits that you pointed out.  I am waiting to merge this PR as I would like to validate that the `index.Tensor_hacked_twin` change works but I am currently blocked by trt-mlir upgrade for testing the `tcp.gather_nd`

PR #103, opened 2024-10-24
[WORK IN PROGRESS] tcp.scatter and tcp.scatter_nd

matthewfl:
This currently includes the tcp.gather_nd PR, so it appears larger than it will be.

There is currently no lowering to linalg for the tcp.scatter/tcp.scatter_nd

More tests are still needed for this

Note: the TosaToSCF.cpp file has an example of tosa.scatter getting lowered to scf, so we could potentially create something similar.


sjain-stanford:
Closing in prep for archiving. Please feel free to reopen against https://github.com/llvm/mlir-tcp.

